PR #51, opened 2018-10-11
README: Add table with learner statuses

pat-s:


pat-s:
> I assume we are doing an mlr-learners org?

That's the current plan.

> Shouldn't be so complicated to get all the repository names from the mlr-learners org and create these links.

Would be great, yes. This is just a preview how it could look like.

PR #55, opened 2018-10-14
[WiP] Roles as list

mllg:
## Problems with the prior approach of putting everything in a `data.table` with columns `id`x`role`x`[...]`:

* Columns may have multiple roles / functions. E.g., a column can be a feature but also an weight. Or the target column determines the required order (for time series tasks).
* (Premature optimization): While `data.table` is really fast for large data, it has quite some overhead optimizing the "queries". Small/atomic operations like "get me the name of the column with role target" quickly add up, making some core operations unnecessary slow.
* The previous point was to some degree mitigated via caching. However, I prefer not having to worry about cache invalidation if not absolutely required.

## The new approach

`row_roles` is a list with slots "use" and "validation", both holding an integer or character of arbitrary length. The setter (`set_row_role`) uses set operations to operate on it.

`col_roles` (not yet implemented) is a list with slots "feature", "target", "order", "weights" (maybe more in the future). It works analogously to `row_roles`. I.e., to remove a feature, you just remove it from the vector `task$col_roles$feature` via `set_col_role`.

Most operations are now much more concise. E.g., no need to inner join a `data.table` to get the vector of feature names, just `task$col_roles$feature`.
W.r.t. computational performance I already reduced the overhead for a simple benchmark using the dummy learner from 2.2s to 0.8s (`attic/profvis.R`). With a bit of refactoring of the data backends I aim to get to < 0.2s.


mllg:
* DataBackendDataTable now uses integer indexing if feasible. 
* `row_info` has been removed in favor of `row_roles`
* `col_info`'s column `role` is now ported to `col_roles`

mllg:
Speedup is ~3x with cached version and ~6x in comparison with uncached version.

mllg:
@berndbischl want to take a look?

PR #56, opened 2018-10-26
Throw a proper error when the package with dataset is missing

zzawadz:
`!nzchar(find.package(package, quiet = TRUE))` returns an empty vector when the `package` is not installed. It causes a non-informative error message in `load_dataset`:

```
 Error in if (!nzchar(find.package(package, quiet = TRUE))) stopf("Please install package '%s' for data set '%s'",  : 
  argument is of length zero 
```

It might be better to change this to:
```
!length(find.package(package, quiet = TRUE))
```


mllg:
Thanks!

PR #64, opened 2018-11-08
Re-enable auto-deployment of docs

pat-s:
### Changes

- pandoc 2.3.1

- devtools -> remotes

- seeds for the vignettes and some examples

### Where are the problems coming from?

The biggest problems are outputs that have a "time" slot because this slot changes in every CI run.

Either we omit this slot in the output or make the corresponding sections "read-only". 

This branch finally contains a working vignette version that only commits if there have been real changes to the docs. See [this build](https://travis-ci.org/mlr-org/mlr3/builds/452144242) as an example.

### Auto-deployment
Also it should be safe now to commit man/, DESCRIPTION and NAMESPACE changes via travis/tic. At least it worked without problems in my latest tests.

So basically we have two auto-deploys in a travis run now: One for the pkgdown site and one for man/, DESCRIPTION and NAMESPACE.

We might need to give Travis a few runs to "incorporate" all the latest changes. Once we see a commit that did not trigger an auto-deploy (if there was no change to the docs) we can merge this here.


pat-s:
@mllg 	I think we're good to go here. Latest build: https://travis-ci.org/mlr-org/mlr3/builds/452375639

Let me know if sth is still unclear.

PR #107, opened 2018-12-21
add check if resample object is intantiated

schalkdaniel:
See issue #108. This would be very neat for nested resampling.


mllg:
As explained in #108, this won't work if multiple tasks are provided.

PR #112, opened 2018-12-24
Add assert_task_type and assert_learner

pat-s:
`assert_task_type` can check filters and learners.

 `assert_learner` is consistent with simple class checking assertions like `assert_task` or `assert_backend`.

cc @Coorsaa 	@QuayAu


mllg:
What is the advantage of keeping `assert_learner()` (and other type-specific assertions) simple rather than extending it with more functionality?

pat-s:
> What is the advantage of keeping assert_learner() (and other type-specific assertions) simple rather than extending it with more functionality?

To keep it simple (isn't that our goal?) and have really descriptive assertions.

PR #139, opened 2019-01-17
autotest all predict types. For classif: test binary and multiclass.

QuayAu:
see https://github.com/mlr-org/mlr3learners/issues/13


QuayAu:
I added some sanity checks: 

classif: very easy 1d separable two class task. tests if mmce < 0.3
regr: constant 1d task: f(x) = 1 for all x. tests if mse < 0.3

tests/testthat/test_mlr_learners_regr_featureless.R
```diff
@@ -1,34 +1,7 @@
...
-test_that("Predict with se", {
-  task = mlr_tasks$get("bh")
-  learner = mlr_learners$get("regr.featureless")
-  learner$predict_type = "se"
-  expect_learner(learner, task)
-
-  e = Experiment$new(task, learner)$train()$predict()
-  expect_numeric(e$data$prediction$se, len = task$nrow, any.missing = FALSE, lower = 0)
-})
-
```
mllg: Why have you removed these tests? Can you add them again?

QuayAu: I thought we wanted to streamline the unit tests with an auto test for each learner. Everythin, what is tested here (simple train predict + predict type "se") is still tested with expect_autotest().

mllg: Right, except these line:
```r
expect_class(e$model, "featureless")
expect_numeric(e$model, len = 2L, any.missing = FALSE)
```
But we can live without these tests.

PR #148, opened 2019-01-28
Learner autotests: improved debugging for learner errors

Coorsaa:


Coorsaa:
Each learner is tested via `expect_autotest(learner)'
If an error occurs, we can simply access the related experiment and error message by 
```
test = expect_autotest(learner)
# experiment
test$Experiment
# error
test$Error
``` 

right now all sanity checks fail

PR #157, opened 2019-02-01
Add Appveyor

pat-s:
Looks stable now.

Not sure about the pandoc error when checking vignettes (might resolve itself). I already tried with a fresh cache but didn't help.
We do not get this error in `mlr` or others, so might be really on our side here.

With caching enabled, a build takes about 5min 30s.
If this works for some days, I'll add the template to the wiki.


pat-s:
Removed the vignette checking and building completely, this is done only on Travis to avoid all the `pandoc` problems.

PR #161, opened 2019-02-05
param_vals are now in param_set

berndbischl:


berndbischl:
this runs with the new version of paradox and passes tests

I have not removed the ABs for param_vals, but rather changed them, in order to minimize conflicts 
but i think they should be removed and some search / replace is in order for this

berndbischl:
i think you also have to remove some docu sections

mllg:
Closed in favor of #168

PR #257, opened 2019-06-24
Update `assert_task()`

pat-s:
- allow empty rows and cols (FALSE by default)
- allow checking for a minimum number of rows and cols

cross-ref https://github.com/mlr-org/mlr3featsel/issues/39


R/assertions.R
```diff
@@ -27,8 +27,18 @@ assert_experiment = function(e, .var.name = vname(e)) {
 #'   Set of allowed feature types.
 #' @param task_properties :: `character()`\cr
 #'   Set of required task properties.
+#' @param allow_empty_rows :: `logical()`\cr
```
mllg: Is this the same as setting `min_rows` to 1?

R/assertions.R
```diff
@@ -27,8 +27,18 @@ assert_experiment = function(e, .var.name = vname(e)) {
...
+#'   Is a task without observations allowed?
+#' @param minimum_nrows :: `integer()`\cr
+#'   Is a minimum amount of observations required?
+#' @param allow_empty_cols :: `logical()`\cr
```
mllg: Is this the same as setting `min_cols` to 1?

R/assertions.R
pat-s: Yes, changed it :+1:

mllg:
Thanks!

PR #266, opened 2019-06-29
use param_set$get_values instead of $params

mb706:
Make use of mlr-org/paradox#236 . Should not be merged before mlr-org/mlr3learners#26.


mllg:
Superseded by #267.

PR #278, opened 2019-07-14
Add example to README

jakob-r:
Can someone check if this is sufficient?

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -11,3 +11,4 @@
 ^attic$
 ^pkgdown$
 ^inst/extdata/.+\.R$
+^README\.Rmd$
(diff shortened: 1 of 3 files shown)


jakob-r:
Preview can be seen in the branch:

pfistfl:
Some suggestions: 
-> A flashy picture in the header would be really cool
-> Why id = cars?
-> Maybe have a subsubsection between train/test and resampling

pfistfl:
The old mlr repo has some more things we might want to add: https://github.com/mlr-org/mlr
- Blog 
- Slack

mllg:
We have a SO hashtag: `"mlr3"`. Can you add this?

jakob-r:
Should be mergable. Other thigs can be done later I guess.

PR #296, opened 2019-08-01
document how to set the lgr level in mlr_control()

pat-s:
relates to https://github.com/mlr-org/mlr3book/issues/41

Will also do it in the book now.


mllg:
I'll add a section about the logger to `resample()` and `benchmark()`.

PR #358, opened 2019-09-23
Check package on circle ci

pat-s:
- builds are running on debian:testing with the latest R version
- deployment is still WIP (maybe we can get rid of Travis in the long term)

- cache needs to be built from scratch every day - this can be automated by a cron build at 3 am in the morning (not needed for mlr3 for runtime reasons)

- packages are installed in parallel with 6 cores

- CPUs are faster than on Travis

- a [build without cache](https://circleci.com/gh/mlr-org/mlr3/2) takes 3:38 min, a [build with cache](https://circleci.com/gh/mlr-org/mlr3/3) takes 2:39

# misc

- Small fix to README


pat-s:
Hm, interesting - one test is constantly failing. The question is: Is Travis not catching it or is it a false positive?

pat-s:
I can reproduce the error locally using an isolated environment with

```r
# remotes::install_github('rstudio/renv')
renv::init():renv::install()
devtools::test()
```

It appears when using the current CRAN version of `mlr3misc`. The error is gone with the dev version of `mlr3misc`.
I suspect we update 'too much' on Travis compared to whats written in the DESCRIPTION file of the package (mlr3 should use CRAN mlr3misc).

However, Travis should do exactly the same as CircleCI since both are running with the same tic template?!

mllg:
The test should be fixed now.

pat-s:
What concerns me more is the fact that our Travis checks are not really using an isolated environment with only the packages and their versions specified in the DESCRIPTION.

I will try to fix that upstream by using _renv_.

mllg:
> What concerns me more is the fact that our Travis checks are not really using an isolated environment with only the packages and their versions specified in the DESCRIPTION.

This is just another case of cache invalidation. Packages in the cache have been installed from GH and hence they are already in the library, travis sees so reason to re-install them from CRAN.

pat-s:
> This is just another case of cache invalidation. Packages in the cache have been installed from GH and hence they are already in the library, travis sees so reason to re-install them from CRAN.

This is why the use of `renv` would really help here since it creates an external library with only the packages from the DESCRIPTION of the package. It does use packages in the general R library if they exists, otherwise the missing deps are installed.

And yes, this library can also be cached then. This should make it a flawless experience. I think this is the way to go.

pat-s:
Merging this now so that the checks on master can pass.

PR #399, opened 2019-11-10
Prototype for R6 Roxygen

mboecker:
As we discussed @jakob-r I tried to convert our mlr3 docs to the R6 Roxygen docs. I'm not sure if "@usage NULL" should still be needed? I included it because otherwise the usage section would just say "LearnerClassif".


mllg:
This is not really converted to the new documentation approach in roxygen2 v7. Have you read https://roxygen2.r-lib.org/articles/rd.html#r6 ?

mboecker:
Thanks for the link I havent seen that documentation. I have worked off the inital issue for R6 support in roxygen (https://github.com/r-lib/roxygen2/pull/922) but since there is actual documentation, I will gladly take another look at it.

mboecker:
the new commit should actually document this using the new R6 syntax. sorry for my misunderstanding. I documented the parameters of $new by itself in this file but they would usually documented as inherited parameters from Learner.

pat-s:
Can you give a quick summary on the current status?

jakob-r:
@mboecker is still working on it.

mllg:
Superseded by #439.

PR #439, opened 2020-02-18
R6 roxygen transition

pat-s:
# Changes

- Group `as_data_backend` S3 methods into `@name as_data_backend`

- add `man-roxygen` files to avoid redundant params


mllg:
I migrated Task, TaskSupervised and TaskClassif. I think there are some points open which are not clear to me:

* How do we annotate the type for params? Put them in parentheses or not?
* Templating is not working very well. If you have multiple `@template rows` in your documentation (e.g., for multiple methods), the documentation for the parameter shows up multiple times in the documentation for each method.
* If you overload methods, you seem to have to document them again, and there seems to be no way to inherit the documentation from the super class?

pat-s:
> How do we annotate the type for params? Put them in parentheses or not?

Don't have a real opinion on this.
Roxygen inserts line breaks anyway after the param name so the "type" will not directly appear after the param name.

Hence I'd say no brackets?

> Templating is not working very well. If you have multiple @template rows in your documentation (e.g., for multiple methods), the documentation for the parameter shows up multiple times in the documentation for each method.

Can't find a problematic example. Which class is affected by this?
`@template` should not behave differently than `@params` AFAIK?

> If you overload methods, you seem to have to document them again, and there seems to be no way to inherit the documentation from the super class?

With overload you mean define an already existing method of the superclass in the subclass? In these cases inheritance won't work but I thinks that's fine as arguments of redefined subclass functions could theoretically deviate from the ones in the superclass.

mllg:
> Don't have a real opinion on this.
> Roxygen inserts line breaks anyway after the param name so the "type" will not directly appear after the param name.
> 
> Hence I'd say no brackets?


Uhm no?

```
#' @param rows `integer()`\cr
#'   Row indices.
```
is translated to:
```
\describe{
\item{\code{rows}}{\code{integer()}\cr
Row indices.}
}
```

So the type annotation is in the same line.



> Can't find a problematic example. Which class is affected by this?

For example `DataBackendDataTable$data()`.


> `@template` should not behave differently than `@params` AFAIK?

Possibly yes, I have trouble with both.


> With overload you mean define an already existing method of the superclass in the subclass? In these cases inheritance won't work but I thinks that's fine as arguments of redefined subclass functions could theoretically deviate from the ones in the superclass.

True, but tedious to document. We now have many places where we need to copy-paste.

pat-s:
> For example DataBackendDataTable$data().

RStudio does some tweaking behind the scenes apparently.

**RStudio**

image

**Terminal**

image



> If you have multiple @template rows in your documentation (e.g., for multiple methods), the documentation for the parameter shows up multiple times in the documentation for each method.

I'd say this is by design. Every method comes with its own arguments and needs to be documented as a standalone functions.

Inheritance only works for methods and fields of the super class - there is no avoidance of redundant params within a class.
I think that's fine?

Also `@template` is nothing more than a template for a specific piece of roxygen documentation. It helps for params with appear multiple times throughout the package. Changing it in the template changes all instances of it - avoiding the danger of missing out on one and reducing c/p hassle.

> True, but tedious to document. We now have many places where we need to copy-paste.

Yes but I actually thinks that's even better. Why? It reduces the risk of some fields/params to get overlooked. 
This happens easily when documentation and code is decoupled.
Now we have to document the params right where they live - and get a warning if there are undocumented ones.

I came across a few instances where there were some params being undocumented - roxygen now protects us from this.

mllg:
> I'd say this is by design. Every method comes with its own arguments and needs to be documented as a standalone functions.

Please roll back to your last commit and render again. You will see that for each method the argument "rows" is listed around 5 times.



> Yes but I actually thinks that's even better. Why? It reduces the risk of some fields/params to get overlooked.
> This happens easily when documentation and code is decoupled.
> Now we have to document the params right where they live - and get a warning if there are undocumented ones.

Templating does not work for methods, only for fields (again, roll back to your commit, check the rendered page ... the methods documented via template are not included in the output).

pat-s:
> Templating does not work for methods, only for fields (again, roll back to your commit, check the rendered page ... the methods documented via template are not included in the output).

Ok saw it now.

So turns out that one only has to use `@template rows` once then per Rscript to apply it in all instances. 
If a certain param should be described uniquely one can still use `@params rows` to take precedence.

I.e. templating is still useful, we should just only apply it once per Rscript?

mllg:
> Ok saw it now.
> 
> So turns out that one only has to use `@template rows` once then per Rscript to apply it in all instances.
> If a certain param should be described uniquely one can still use `@params rows` to take precedence.
> 
> I.e. templating is still useful, we should just only apply it once per Rscript?

This is what I've done in the last commits. It is a bit contrary to the concept of keeping code and documentation together, but better than copy-paste.

pat-s:
> This is what I've done in the last commits. It is a bit contrary to the concept of keeping code and documentation together, but better than copy-paste.

Agree. We should change this per repo - for ml3, where there is a lot of argument repetition within one Rscript, this helps. For other repos we might want to document single params in the respective functions.

PR #453, opened 2020-03-06
set task_nrow in custom resampling (Fixes #451)

### LINKED ISSUE #451: custom resampling doesn't set task_nrow
Working with @pfistfl and we found this bug when creating a new resampling object. It appears that `$task_nrow` is not set when using `rsmp("custom")`.

```
library(mlr3)
task = tsk("iris")
resampling = rsmp("custom")
resampling$task_nrow
resampling$instantiate(task,
    train = list(1:110),
    test = list(111:150)
)
resampling$task_nrow
## [1] NA
design = benchmark_grid(
    tasks = task,
    learners = learner,
    resamplings = resampling
)
## Error in benchmark_grid(tasks = task, learners = learner, resamplings = resampling) : 
##  Resampling is instantiated for a task with a different number of observations
```

jakob-r:
Adresses the bug #451.
Apparently `self$task_nrow` was not set.

I think it is not optimal to not call `super$instantiate()` because all changes on `Resample$instantiate()` have to be copied here.
But `Resampling$instantiate()` relies on `private$.sample()` which we than had to *fake* for `ResamplingCustom` which also sounds weird.

### DIFF
--- a/R/ResamplingCustom.R
+++ b/R/ResamplingCustom.R
@@ -51,6 +51,7 @@ ResamplingCustom = R6Class("ResamplingCustom", inherit = Resampling,
       assert_list(test_sets, types = "atomicvector", len = length(train_sets), any.missing = FALSE, null.ok = TRUE)
       self$instance = list(train = train_sets, test = test_sets)
       self$task_hash = task$hash
+      self$task_nrow = task$nrow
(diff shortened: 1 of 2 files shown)


mllg:
Thanks

Issue #119, opened 2019-01-01
Add support for downloading openml data sets


mllg:
Would be great for examples.

Better suited for the openml package, but you would get sooo much with just these lines:

```r
  id = 13
  j = jsonlite::fromJSON(sprintf("https://www.openml.org/d/%i/json", id))
  fn = tempfile()
  download.file(j$url, fn)
  data = farff::readARFF(fn)
  target = j$default_target_attribute
```

berndbischl:
+1
but we need to discuss this a bit

berndbischl:
OTOH: if its just a few lines of code, it might be a good idea to just go for it and play around with it and see what happens?

berndbischl:
i really dont think this should be in mlr3. close?

pat-s:
I also do not think that this should live in {mlr3}.
The issue here is {farff} with what it would pull in ({stringi}, {readr}).
{jsonlite} is for free.

How about putting this code as an example into a code chunk within a section of the help page of `Task`?

berndbischl:
IMHO this should simply be closed.

pat-s:
Probably. Presumably this code might break in the future and having it read-only in a help page might result in undetected dead code.

PR #485, opened 2020-04-12
openml integration

mb706:
Could be its own package but that would probably look similar to this. Maybe someone (a hiwi?) wants to go on working with this.
Inspired by #119


mb706:
```r
> ts = tsk("openml", 261)
trying URL 'https://www.openml.org/data/v1/download/31/credit-g.arff'
Content type 'text/plain;charset=UTF-8' length 162291 bytes (158 KB)
==================================================
downloaded 158 KB

Parse with reader=readr : /tmp/Rtmp41jsuQ/file35862627d240
header: 0.025000; preproc: 0.000000; data: 0.005000; postproc: 0.001000; total: 0.031000
> rs = rsmp("openml", 261)
trying URL 'https://www.openml.org/api_splits/get/261/Task_261_splits.arff'
Content type 'text/plain;charset=UTF-8' length 13697 bytes (13 KB)
==================================================
downloaded 13 KB

Parse with reader=readr : /tmp/Rtmp41jsuQ/file35867427818f
header: 0.001000; preproc: 0.001000; data: 0.003000; postproc: 0.000000; total: 0.005000
> resample(ts, lrn("classif.rpart"), rs)$aggregate()
INFO  [12:52:17.639] Applying learner 'classif.rpart' on task 'openml_credit-g' (iter 1/1) 
classif.ce 
 0.2909091 
```

mb706:
Currently missing:
* caching
* giving resampling an intelligent name from OpenML
* possibly enabling `rsmp("openml")$instantiate(task)` if we save the OpenML Task ID in `task`
* handling task types beyond classification and regression
* using a different farff reader
* putting jsonlite and the farff reader(s) into `Suggests`

mb706:
* controlling verbosity (through logger I presume)
* check if the resampling fold metadata really alsays has the format that I am assuming
* `.combine` function should be possible for resampling in almost all cases

mllg:
There is also https://github.com/mlr-org/mlr3oml/tree/testing. 

And I would like to have it in an extra package I guess. Without the extra dependencies it is not usable.

mb706:
Hm wasn't aware of the state of mlr3oml. I think I like the `tsk("openml", ...)` interface I show here, maybe some of the ideas here can be ported over there.

mllg:
I've updated https://github.com/mlr-org/mlr3oml and ported the dictionary integration.

PR #514, opened 2020-06-25
replace german_credit data with corrected data from rchallenge package

sumny:


sumny:
> Thanks! Will this affect posts in mlr3gallery?

In principle, this affects:

* 2020-03-11-basics-german-credit (runs fine because no specific use of the internal structure of variables is made here)

* 2020-03-11-mlr3pipelines-tutorial-german-credit (should run fine but is no longer very sensible because now there are only 3 integer features being used and the filter is using 3 features, therefore needs some minor changes)

* 2020-03-11-mlr3tuning-tutorial-german-credit (runs fine but uses data from here https://github.com/mlr-org/mlr-outreach/raw/master/2019_madrid/mlr3tuning/randomsearch_3600.rds which uses the old `german_credit`; not sure whether other outreach files need to be changed/looked at as well)

* 2020-03-30-stratification-blocking (runs fine, no changes needed)

* 2020-01-31-encode-factors-for-xgboost (needs minor wording changes because there is now more than one ordered factor and `unemployment_duration` is now a factor)

pat-s:
> I could open an issue in mlr3gallery and fix the posts above?

Yes, please.

Thanks!

mllg:
Thanks. I'll merge as soon as there are fixes for the gallery.

mllg:
Thanks, will merge now, upload, and repair the gallery posts.

Issue #528, opened 2020-07-27
convert_task and convert_prediction helpers


pfistfl:
Adding on to a recent call w/ @mllg, @pat-s  it would perhaps be interesting to not only have
a [convert_task](https://github.com/mlr-org/mlr3pipelines/blob/master/R/convert_task.R) function but analogously a `convert_prediction` function. See https://github.com/mlr-org/mlr3proba/issues/141 for example. 

Additionally, for `convert_task`, we should check whether our function can handle the cases required in `mlr3proba`.

mllg:
PR would be welcome.

mllg:
We have some support for converting tasks, not clear how this would be defined for predictions.

If still needed, just re-open.

PR #529, opened 2020-07-27
Task converters

### LINKED ISSUE #470: task constructors and re-instantiating tasks
- This is not directly related to `mlr3spatiotempcv`, but the first package where this occurs, so I'll try to discuss this here. While developing `mlr3pipelines` I came across the following issue.
If I, for example, want to set a different `target` for a given task, I would basically have to reconstruct the `Task`. Currently, I construct a task of `new_type` as follows:
```
  tsk = get(mlr_reflections$task_types[mlr_reflections$task_types$type == new_type, ]$task)
  newtsk = tsk$new(id = intask$id, backend = intask$backend, target = new_target)
  # copy col_roles & row_roles
```

This does not work with  `mlr3spatiotempcv` as `new_type` is now `classif` and constructs a `Classif` task but the task should be a `ClassifST` task. In addition additional fields for `TaskClassifST` (`crs`, `coordinate_names`)  are required.

It would also be very cool if I could just copy a fixed set of other fields in order to obtain a valid re-instantiation of a task.

Would be possible if we made the following changes:
- `coordinate_names` is an active binding, setting / getting the corresponding `col_role`.
- `crs` is an active binding to a `$crs` item in a list called `task_metadata` (this should be in `mlr3`).

I could then instantiate a new `TaskClassifST`, copy col_roles, row_roles and task_metadata and would then have a valid copy of the original task.
  

  

- Just FYI: `self$crs` is added twice in [TaskClassifST](https://github.com/mlr-org/mlr3spatiotempcv/blob/master/R/TaskClassifST.R)

mllg:
Fixes #470, and partly also #528.


mllg:
* Tasks have new field `extra_args` to store additional constructor arguments.
* Copied `convert_task()` from mlr3pipelines 
* Added convenience functions `as_regr_task()` and `as_classif_task()`

mllg:
TaskClassif is unchanged, so nothing should break.

mllg:
So @RaphaelS1 should check if `convert_task()` works for him, right? Additionally, there should be an implementation of `as_task_surv()`, `as_task_surv.TaskRegr()` and `as_task_regr.TaskRegr()` in mlr3proba?

RaphaelS1:
Sorry jumping into this quite late so just catching up on the additions. I'm slightly confused by this feature, would these functions be called by a user or in a pipeline, or both?

The current implementation wouldn't really make sense to be used in proba because, e.g. for TaskRegr -> TaskSurv either a censoring column is automatically added that assumes no-one is censored, but in this case no one would use survival models, or there has to be a way for the user to add in a second task (or just a vector) with the censoring column. In the reverse, TaskSurv -> TaskRegr, there are multiple methods for doing this and some involve multiple tasks and additional parameters. These would never work therefore outside of a pipeline. Similarly `TaskSurv -> TaskClassif` could maybe make sense but generally this transformation is usually seen to multi-label classification.

I like the feature but I'm just saying that if this is *instead of* a pipeline then I'd have to additionally include task transformer pipelines in proba and as it is it's slightly too rigid to make any practical sense for a survival. Also really like dispatch on data.frames, that will be a big help.

mllg:
Thanks for your feedback Raphael. AFAICT the implementation has the following purposes:

1. Provide a simple solution for task conversion for the user. E.g., if the user wants to impute a numeric column of a classification task with a regression learner. This does not necessarily mean that the conversion has to be free of arguments.
2. Reduce the complexity of code in PipeOps. Again, it is okay if arguments are required to do the conversion.

pat-s:
> TaskClassif is unchanged, so nothing should break.

I see now, the constructor remains unchanged. This is cool and means I do not need to make additional changes in spatiotempcv - only store the constructor args in a list.

RaphaelS1:
> 1. Provide a simple solution for task conversion for the user. E.g., if the user wants to impute a numeric column of a classification task with a regression learner. This does not necessarily mean that the conversion has to be free of arguments.
> 2. Reduce the complexity of code in PipeOps. Again, it is okay if arguments are required to do the conversion. Also note that the DataBackend is not copied during conversion. Picking up the example from (1), you could convert a TaskClassif to TaskRegr, call `learner$predict()` to generate imputed values, convert the task back to a classification task and `task$cbind()` the imputed column. All without copying the data around.

Ah sorry I had missed that! Sounds great and really useful. I still think there should be a pipeop that does this too, even if it just calls the function internally, because there are use cases for graph learners to transform a task and others to use the original, it wouldn't be possible to benchmark these if conversion is only possible outside of pipelines.

pfistfl:
> I still think there should be a pipeop that does this too
Agree

So in `TaskSurv` -> `TaskRegr` you would keep the censoring columns in the data, but basically just switch the `col_roles`?
If I understand things correctly, i.e. in surv you have censoring columns as a `col_role`, and this info would just be deleted when converting to a `TaskRegr`.  When converting back, you would basically just need to set those `col_roles` again and you'd be good?

For some other reduction strategies, e.g. `multiclass_to_binary`, `ordinal` and `multilabel` we will also have to create `PipeOp` s that pipe around the original factor levels etc. I guess we should talk about a good abstraction principle here at some point, @RaphaelS1 might have additional insights!

Also pinging @mb706 for some thoughts, have we missed anything?

RaphaelS1:
> So in TaskSurv -> TaskRegr you would keep the censoring columns in the data, but basically just switch the col_roles?
If I understand things correctly, i.e. in surv you have censoring columns as a col_role, and this col_role would just be deleted when converting to a TaskRegr. When converting back, you would basically just need to set those col_roles again and you'd be good?

Censoring isn't set as a col_role but as a target, i.e. the target is formed of two columns and not one. And dropping this for TaskRegr is one strategy out of many (imputation, omission, deletion,.....). Converting back would require supplying censoring as a new column because it's not something you can reorder or keep in the TaskRegr.

> For some other reduction strategies, e.g. multiclass_to_binary, ordinal and multilabel we will also have to create PipeOp s that pipe around the original factor levels etc. I guess we should talk about a good abstraction principle here at some point, @RaphaelS1 might have additional insights!

Definitely interested in discussing multi-label.

pfistfl:
Perhaps export `convert_task`? At least `mlr3multioutput` could use it.

pat-s:
@mllg I think `TaskClassif` and `TaskRegr` also need `extra_args` in the constructor as I am calling `super$initialize` from `TaskClassifST` with `TaskClassif` as the super class.

pat-s:
@mllg Ready to merge?

I also adapted {mlr3spatiotempcv} to the changes here: https://github.com/mlr-org/mlr3spatiotempcv/pull/64.

PR #550, opened 2020-09-22
Add use_future flag

be-marc:
Closes https://github.com/mlr-org/mlr3tuning/issues/270

* If the `future` package is installed, `resample()` and `benchmark()` call `future.apply::future_mapply`  even if `future::plan(sequential)` is set.
* Can be disabled by setting `use_future = FALSE` in `resample()` and `benchmark()` 
* Allows parallelization of the inner loop in nested resampling with `future::plan(list(sequential, multisession))`


be-marc:
Unit test fails because PIDs are duplicated in parallel mode on gh actions. Locally PIDs are unique.

mllg:
I'll just remove the possibility to disable future for now.

PR #554, opened 2020-09-28
retrain

### LINKED ISSUE #272: learner API interface: update_internal
allow an online update function. the contract should be that
the model slot is set, and n new data points are passed. 

an online trainable model should now have a (fast) update function available

we also need a property "online_update" 

another option would be to handle this in train_internal, and implement a case that the model should be updated if the model slot is set and not NULL. but these might be confusing and error prone

be-marc:
Closes #272 and https://github.com/mlr-org/mlr3learners/issues/7

## Retrain

Extends learner API with a `$retrain()` method which continues the training of models e.g. continue the training of xgboost with additional boosting iterations.   

```
library(mlr3learners) # @continue branch

task = tsk("iris")
learner = lrn("classif.xgboost", nrounds = 10) 
learner$train(task)

learner$retrain(task, list(nrounds = 20))
```

To use continuable models during tuning, `ResampleResult` and `BenchmarkResult` implement `$retrain()` methods.

```
task = tsk("iris")
learner = lrn("classif.xgboost", nrounds = 10) 
rr = resample(task, learner, resampling, store_models = TRUE)

rr$retrain(list(nrounds = 20), store_models = TRUE)
```

```
task = tsk("iris")
learner1 = lrn("classif.xgboost", nrounds = 10, eta = 0.1) 
learner2 = lrn("classif.xgboost", nrounds = 10, eta = 0.5) 
learners = list(learner1, learner2)
design = benchmark_grid(task, learners, resampling)
bmr = benchmark(design, store_models = TRUE)

bmr$retrain(list(nrounds = 20), store_models = TRUE)
```


R/worker.R
```diff
@@ -81,6 +80,131 @@ learner_train = function(learner, task, row_ids = NULL) {
...
+      log = rbindlist(list(learner$state$log,
+        append_log(NULL, "train", result$log$class, result$log$msg))),
+      train_time = learner$state$train_time + result$elapsed
+    ))
+  }
+
+  learner
+}
+
+learner_update = function(learner, task, row_ids = NULL) {
```
jakob-r, line 146: `learner_update` looks like the exact same function as `learner_continue` and both look pretty much like 
Lets have one function `learner_train(learner, task, row_ids, mode = c("train", "update", "continue"))`

jakob-r (commented):
And maybe you can already write one simple example within a unit test.

be-marc:
> And maybe you can already write one simple example within a unit test. 

The unit tests are currently in mlr3learners because using a learner from mlr3learners in mlr3 causes other unrelated unit tests to fail.

jakob-r:
> The unit tests are currently in mlr3learners because using a learner from mlr3learners in mlr3 causes other unrelated unit tests to fail.

That I just realized. We can implement `$continue` and `$update` for `Learner*Debug`.
Implementing `$update` for featureless is possible but makes them a tiny bit more complex - the model has to store the number of observations (per class) in the model.
Maybe vote with :+1: / :-1: whether we want to have `$update` for `Learner*Featureless`.

jakob-r:
Looks much nicer now :)

jakob-r:
I have not looked in-depth but the rewrite looks clean and well so far.

be-marc:
moved to new PR

PR #556, opened 2020-09-29
Add timeout field to learners

mllg:
Requires mlr3misc-0.6.0.


mllg:
Fixes #556.

jakob-r:
This only has effect if `$encapsulate = c(train = "callr", predict = "callr")` is set (or similar). right? Or does it always work? It somehow feels like it should be an encapsulation parameter and not a  Parameter of the `Learner`-Object. Then we could also have memory limits etc following a similar API?

mllg:
> This only has effect if `$encapsulate = c(train = "callr", predict = "callr")` is set (or similar). right?

Also works with `"none"` and `"encapsulate"`, but only during user interrupts. 

> It somehow feels like it should be an encapsulation parameter and not a Parameter of the `Learner`-Object. Then we could also have memory limits etc following a similar API?

Agreed, but I couldn't come up with more options. Limiting memory is not really possible with R.

PR #575, opened 2020-11-05
Unsupervised Task

damirpolat:
This is the first draft for TaskUnsupervised. The main difference right now is that I removed `target` value from `col_roles` field in TaskUnsupervised . I also moved `formula` from Task into TaskSupervised since it's not relevant for unsupervised tasks. 

Thanks, 
Damir

### DIFF
--- a/R/Task.R
+++ b/R/Task.R
@@ -163,16 +163,6 @@ Task = R6Class("Task",
     },
 
     #' @description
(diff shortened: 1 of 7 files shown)


mllg:
Actually, I'm not sure why mlr3 needs to know about cluster tasks. Maybe we can setup a video call or discuss this in mattermost.

PR #589, opened 2020-12-16
Add list_mlr_tasks.R

RaphaelS1:
Adds a function for listing tasks in mlr3.

I'm guessing you probably don't want a separate function so feel free to strip out any useful code and merge into `as.data.table(mlr_tasks)`.


mllg:
I've added a properties column.

PR #592, opened 2021-01-15
Image datatype

pfistfl:
Adds a datatype `imagepath`, which allows to specify a path to an image as a covariate.


mllg:
Implemented in bbdd90a; this is now a bit more general. Instead of a special data type "image", you can just use a character column and give it the role "uri" (query with `task$uri`). I don't see any downsides, but let me know if this causes problems on your end.

PR #603, opened 2021-01-30
Iris -> penguins

RaphaelS1:
As discussed, iris dataset is changed to penguins in all user-facing documentation and examples. Tests are unchanged.


mllg:
Thanks!

PR #621, opened 2021-03-22
test with paradox performance improvement

mb706:


pat-s:
Is this PR still being worked on?

mllg:
I guess keeping this just as a branch would also be fine.

PR #622, opened 2021-03-29
Support retrain for learners

### LINKED ISSUE #272: learner API interface: update_internal

be-marc:
Closes #272 and https://github.com/mlr-org/mlr3learners/issues/7

## retrain a Learner

Allows to retrain a learner e.g. continue the training of a `xgboost` model with additional boosting iterations or reduce the number of trees in a `ranger` model.   

`learner$retrain(task, param_vals)` retrains a learner.

```r
library(mlr3learners) # @continue branch

task = tsk("iris")
learner = lrn("classif.xgboost", nrounds = 10) 
learner$train(task)

learner$retrain(task, list(nrounds = 20))
```

`learner$is_retrainable(param_vals)` checks if learner can be retrained with hyperparameter values in `param_vals`.

```r
library(mlr3learners) # @continue branch

task = tsk("iris")
learner = lrn("classif.xgboost", nrounds = 10) 
learner$train(task)

learner$is_retrainable(list(nrounds = 20))
# > [1] TRUE
```

## retrain with resample results

```r
task = tsk("iris")
learner = lrn("classif.xgboost", nrounds = 10) 
rr = resample(task, learner, resampling, store_models = TRUE)

rr$retrain(list(nrounds = 20), store_models = TRUE)
```


be-marc:
Moved to #709

PR #626, opened 2021-04-12
Style package with new {mlr.styler} package

pat-s:
This is a first diff of the new {[mlr.styler](https://github.com/mlr-org/styler.mlr)} package (still private) by @lorenzwalthert.
Lorenz put a [lot of time](https://github.com/mlr-org/styler.mlr/pull/1) into our mlr-style, not only by creating this new package but also in the past.

It should make updating + maintenance more easy and should also serve as an exemplary package for additional styles next to the `tidyverse_style`.
Beforehand we used a [fork of mine](https://github.com/pat-s/styler/tree/mlr-style).

@mllg I am not sure if you use {styler} in practice for {mlr3} but your input on specific re-occurring patterns which are changed in the diff is appreciated.
Remember that we can create the style guide to be passive/non-invasive so that in some cases multiple solutions are allowed / not force styled.

Some years have passed since we evaluated the initial guide. If some of your coding patterns have changed since there, it would now be a good time to discuss these here.

Also cc @berndbischl @mb706 @be-marc @pfistfl @jakob-r @RaphaelS1 (your input is appreciated/desired as well of course as the style applies to all mlr-org repos. {mlr3} is just the demo-repo for now.).

@lorenzwalthert
If there are any new transformers that you know of which would be possibly helpful here, just suggest them :)

### DIFF
--- a/R/BenchmarkResult.R
+++ b/R/BenchmarkResult.R
@@ -270,7 +270,7 @@ BenchmarkResult = R6Class("BenchmarkResult",
       tab = insert_named(tab, scores)
 
       cns = c("uhash", "nr", "resample_result", "task_id", "learner_id", "resampling_id", "iters",
-          "warnings", "errors", "params", ids(measures))
+        "warnings", "errors", "params", ids(measures))
       cns = intersect(cns, names(tab))
       tab[, cns, with = FALSE]
     },
@@ -302,10 +302,11 @@ BenchmarkResult = R6Class("BenchmarkResult",
       learner_phashes = NULL
 
       filter_if_not_null = function(column, hashes) {
-        if (is.null(hashes))
+        if (is.null(hashes)) {
           fact
-        else
+        } else {
           fact[unique(hashes), on = column, nomatch = NULL]
+        }
       }
 
 
--- a/R/LearnerClassifDebug.R
+++ b/R/LearnerClassifDebug.R
@@ -152,8 +152,13 @@ LearnerClassifDebug = R6Class("LearnerClassifDebug", inherit = LearnerClassif,
         if (!is.null(pv$predict_missing)) {
           ii = sample.int(n, n * pv$predict_missing)
           prob = switch(missing_type,
-            "na" = { prob[ii, ] = NA_real_; prob },
-            "omit" = { prob[ii,, drop = FALSE] }
+            "na" = {
+              prob[ii, ] = NA_real_
+              prob
+            },
+            "omit" = {
+              prob[ii, , drop = FALSE]
+            }
           )
         }
       }
--- a/R/TaskClassif_penguins.R
+++ b/R/TaskClassif_penguins.R
@@ -35,7 +35,7 @@ load_task_penguins = function() {
   penguins = as.data.table(palmerpenguins::penguins)
   setnames(penguins,
     old = c("bill_length_mm", "bill_depth_mm", "flipper_length_mm", "body_mass_g"),
-    new = c("bill_length",    "bill_depth",    "flipper_length",    "body_mass")
+    new = c("bill_length", "bill_depth", "flipper_length", "body_mass")
   )
 
   b = as_data_backend(penguins)
--- a/R/bibentries.R
+++ b/R/bibentries.R
@@ -15,15 +15,15 @@ bibentries = c(
 
   bischl_2012 = bibentry("article",
     title = "Resampling Methods for Meta-Model Validation with Recommendations for Evolutionary Computation",
-    author       = "Bernd Bischl and Olaf Mersmann and Heike Trautmann and Claus Weihs",
-    year         = "2012",
-    month        = "jun",
-    journal      = "Evolutionary Computation",
-    publisher    = "MIT Press - Journals",
-    volume       = "20",
-    number       = "2",
-    pages        = "249--275",
-    doi          = "10.1162/evco_a_00069"
+    author = "Bernd Bischl and Olaf Mersmann and Heike Trautmann and Claus Weihs",
+    year = "2012",
+    month = "jun",
+    journal = "Evolutionary Computation",
+    publisher = "MIT Press - Journals",
+    volume = "20",
+    number = "2",
+    pages = "249--275",
+    doi = "10.1162/evco_a_00069"
   ),
 
   breiman_1984 = bibentry("book",
--- a/R/Learner.R
+++ b/R/Learner.R
@@ -244,8 +244,9 @@ Learner = R6Class("Learner",
       }
 
       pdata = learner_predict(self, task, row_ids)
-      if (is.null(pdata))
+      if (is.null(pdata)) {
         return(NULL)
+      }
 
       as_prediction(check_prediction_data(pdata))
     },
@@ -269,8 +270,9 @@ Learner = R6Class("Learner",
       newdata = as.data.table(assert_data_frame(newdata, min.rows = 1L))
 
       if (is.null(task)) {
-        if (is.null(self$state$train_task))
+        if (is.null(self$state$train_task)) {
           stopf("No task stored, and no task provided")
+        }
         task = self$state$train_task$clone()
       } else {
         task = assert_task(as_task(task, clone = TRUE))
@@ -414,7 +416,10 @@ Learner = R6Class("Learner",
       switch(name,
         .param_set = value$clone(deep = TRUE),
         fallback = if (is.null(value)) NULL else value$clone(deep = TRUE),
-        state = { value$log = copy(value$log); value },
+        state = {
+          value$log = copy(value$log)
+          value
+        },
         value
       )
     }
--- a/R/MeasureDebug.R
+++ b/R/MeasureDebug.R
@@ -51,8 +51,9 @@ MeasureDebug = R6Class("MeasureDebug",
   private = list(
     .score = function(prediction, ...) {
       na_ratio = self$param_set$get_values()$na_ratio
-      if (na_ratio > runif(1L))
+      if (na_ratio > runif(1L)) {
         return(NA_integer_)
+      }
       length(prediction$row_ids)
     }
   )
--- a/R/PredictionDataClassif.R
+++ b/R/PredictionDataClassif.R
@@ -90,7 +90,7 @@ c.PredictionDataClassif = function(..., keep_duplicates = TRUE) {
   if (!keep_duplicates) {
     keep = !duplicated(tab, by = "row_ids", fromLast = TRUE)
     tab = tab[keep]
-    prob = prob[keep,, drop = FALSE]
+    prob = prob[keep, , drop = FALSE]
   }
 
   result = as.list(tab)
--- a/R/ResampleResult.R
+++ b/R/ResampleResult.R
@@ -138,7 +138,7 @@ ResampleResult = R6Class("ResampleResult",
         set(tab, j = "learner_id", value = ids(tab[["learner"]]))
         set(tab, j = "resampling_id", value = ids(tab[["resampling"]]))
         setcolorder(tab, c("task", "task_id", "learner", "learner_id", "resampling", "resampling_id",
-            "iteration", "prediction"))
+          "iteration", "prediction"))
       }
 
       if (conditions) {
@@ -213,8 +213,9 @@ ResampleResult = R6Class("ResampleResult",
     task = function(rhs) {
       assert_ro_binding(rhs)
       tab = private$.data$tasks(private$.view)
-      if (nrow(tab) == 0L)
+      if (nrow(tab) == 0L) {
         return(NULL)
+      }
       tab$task[[1L]]
     },
 
@@ -224,8 +225,9 @@ ResampleResult = R6Class("ResampleResult",
     learner = function(rhs) {
       assert_ro_binding(rhs)
       tab = private$.data$learners(private$.view, states = FALSE)
-      if (nrow(tab) == 0L)
+      if (nrow(tab) == 0L) {
         return(NULL)
+      }
       tab$learner[[1L]]
     },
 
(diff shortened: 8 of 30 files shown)


R/BenchmarkResult.R
```diff
@@ -359,9 +366,7 @@ BenchmarkResult = R6Class("BenchmarkResult",
       }
 
       ResampleResult$new(self$data, view = needle)
-    }
-  ),
-
+    }),
```
pat-s, line 369: @mllg 

IIRC we discussed about this pattern/change the last time already and agreed to allow both options?

R/DataBackend.R
```diff
@@ -26,7 +26,8 @@
 #'
 #' @export
 #' @examples
-#' data = data.table::data.table(id = 1:5, x = runif(5),
+#' data = data.table::data.table(
+#'   id = 1:5, x = runif(5),
```
pat-s, line 30: @mllg Is this something we want to have?

R/LearnerClassifDebug.R
```diff
@@ -57,29 +58,28 @@ LearnerClassifDebug = R6Class("LearnerClassifDebug", inherit = LearnerClassif,
...
-          error_predict        = p_dbl(0, 1, default = 0, tags = "predict"),
-          error_train          = p_dbl(0, 1, default = 0, tags = "train"),
-          message_predict      = p_dbl(0, 1, default = 0, tags = "predict"),
-          message_train        = p_dbl(0, 1, default = 0, tags = "train"),
-          predict_missing      = p_dbl(0, 1, default = 0, tags = "predict"),
+          error_predict = p_dbl(0, 1, default = 0, tags = "predict"),
+          error_train = p_dbl(0, 1, default = 0, tags = "train"),
+          message_predict = p_dbl(0, 1, default = 0, tags = "predict"),
+          message_train = p_dbl(0, 1, default = 0, tags = "train"),
+          predict_missing = p_dbl(0, 1, default = 0, tags = "predict"),
```
pat-s, line 65: @lorenzwalthert 

IIRC we had support for this in the old fork - was there a specific transformer or did we do a hacky island solution here?

R/BenchmarkResult.R
pat-s: Tested this on another repo and in our old mlr-style this was not changed by {styler} -> @lorenzwalthert can we remove this styling? :)

lorenzwalthert:
@pat-s thanks, will look into it tomorrow. Alternatively, you could make one PR per scope, starting with `spaces`, instead of all of it together. Maybe it’s easier to review.

R/LearnerClassifDebug.R
lorenzwalthert: alignment is supported in general with {styler}, but only with right-alignment:

``` r
styler.mlr::style_text(
  'call(
  x  =   2, 
  uz = 333
  )
  '
)
#> call(
#>   x  =   2,
#>   uz = 333
#> )
```

<sup>Created on 2021-04-13 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

R/BenchmarkResult.R
lorenzwalthert: Indeed, there was a transformer for that,but for some reason, it was commented out in the revision of your fork I used when c/p the transformers. Adressed in https://github.com/mlr-org/styler.mlr/pull/7.

R/DataBackend.R
lorenzwalthert: You did not have this previously. Fix suggested in https://github.com/mlr-org/styler.mlr/pull/9/files.

R/ResamplingBootstrap.R
```diff
@@ -70,19 +67,15 @@ ResamplingBootstrap = R6Class("ResamplingBootstrap", inherit = Resampling,
       rownames(M) = NULL
       list(row_ids = ids, M = M)
     },
-
```
lorenzwalthert, line 73: I note there are a lot of these line breaks removed with `strict = FALSE`. This is due to `styler:::remove_line_break_in_fun_call()`, which removes empty lines in function calls, but only if no comment is around, as discussed in https://github.com/r-lib/styler/issues/632 and https://github.com/r-lib/styler/issues/627. We can either remove this rule or set a minimal number of lines that are required in order to assume this is a *big call* and the line breaks are used to *group* different arguments (similar to empty lines we insert at the top of a function declaration if the body of the function has more than k lines) and keep the blank lines for these situations.

mllg: Extra newlines normally serve a purpose, I would also opt to disable this.

R/BenchmarkResult.R
```diff
@@ -215,6 +218,7 @@ BenchmarkResult = R6Class("BenchmarkResult",
     #'
     #' @return [data.table::data.table()].
     aggregate = function(measures = NULL, ids = TRUE, uhashes = FALSE, params = FALSE, conditions = FALSE) {
+
```
mllg, line 221: I'm not a big fan of enforcing a newline here. Any reason to keep this?

lorenzwalthert: I can't find a reference anymore in the style guide but at some point the idea was to leave a blank line when the function declaration is long. Assuming you have a function declaration somewhere that has a blank line at the top, would you want to remove it or simply not touch it? If we don't touch it, we also can't touch 
```r
{ 
  # code 
}
``` 
easily (because of the inner workings of styler). A compromise would be to allow up either one or zero blank lines in any of these two curly brace expressions, but not more.

R/LearnerClassifDebug.R
lorenzwalthert: I suggest to manually align them to one of the supported alignments in https://styler.r-lib.org/articles/detect-alignment.html to make sure styler does not *destroy* the formatting of deliberately aligned code.

RaphaelS1:
Just a short comment for now. Points 11 and 17 in the mlr style guide seem to have evolved to the opposite so `return` statements are rare and we decided to explicitly use `::` for most packages except mlr3 and paradox (e.g. we tend to explicitly write `mlr3misc::`)

R/LearnerClassifDebug.R
lorenzwalthert: Left alignment support on the way: https://github.com/r-lib/styler/pull/774

R/ResamplingBootstrap.R
lorenzwalthert: Addressed in https://github.com/mlr-org/styler.mlr/issues/10.

R/BenchmarkResult.R
pat-s: Is this a possible misunderstanding?

I meant to say that we want to allow the 

```
    }
  )
``` 

pattern. And avoid a styling into `})`.
I think this is why I commented out this transformer at some point.

I see that in the pull you explicitly commented it in again.
Assuming that it was commented out in this current draft - why did this styling happen then? 🤔 
Or am I on the wrong path?

R/DataBackend.R
pat-s: Thanks 🎉

R/LearnerClassifDebug.R
pat-s: great 🎉

R/ResamplingBootstrap.R
pat-s: 🎉

R/BenchmarkResult.R
pat-s: > I can't find a reference anymore in the style guide

The style guide might also not be 100% up-to-date 😅 

I think being deliberate here is probably best - i.e. allowing both.

> A compromise would be to allow up either one or zero blank lines in any of these two curly brace expressions, but not more.

Sounds like a good compromise to me (if that is easier for {styler} also). @mllg?

I think we never want to have two blank lines anywhere and any of zero or one is fine?

pat-s:
> Just a short comment for now. Points 11 and 17 in the mlr style guide seem to have evolved to the opposite so return statements are rare and we decided to explicitly use :: for most packages except mlr3 and paradox (e.g. we tend to explicitly write mlr3misc::)

Yeah, the guide is probably not up-to-date or followed fully in practice.
I think {styler} does not touch semantic things like adding/removing `return()` statements IIRC.

The same goes for explicit namespacing I think - this is probably also not trivial and would require quite some magic to infer this from the function name.

But yeah, we should probably adjust this in the guide.

R/BenchmarkResult.R
lorenzwalthert: Yup, there was a misunderstanding (and also merge conflicts). We want to activate the rule that adds a line break if there is none and remove the rule that removes the line break. Force pushed to https://github.com/mlr-org/styler.mlr after rebasing on main.

R/LearnerClassifDebug.R
lorenzwalthert: You need to have the devel version of styler with `9f9bdff` at least yes. I am adding a minimal version requirement to {styler.mlr} aud bump the devel version in {styler}.

lorenzwalthert: Done. I suggest to re-style.

R/TaskClassif_penguins.R
```diff
@@ -35,7 +35,7 @@ load_task_penguins = function() {
   penguins = as.data.table(palmerpenguins::penguins)
   setnames(penguins,
     old = c("bill_length_mm", "bill_depth_mm", "flipper_length_mm", "body_mass_g"),
-    new = c("bill_length",    "bill_depth",    "flipper_length",    "body_mass")
+    new = c("bill_length", "bill_depth", "flipper_length", "body_mass")
```
pat-s, line 38: @lorenzwalthert DO you have an idea why this alignment is not honoured?

R/bibentries.R
```diff
@@ -15,15 +15,15 @@ bibentries = c(
...
-    doi          = "10.1162/evco_a_00069"
+    author = "Bernd Bischl and Olaf Mersmann and Heike Trautmann and Claus Weihs",
+    year = "2012",
+    month = "jun",
+    journal = "Evolutionary Computation",
+    publisher = "MIT Press - Journals",
+    volume = "20",
+    number = "2",
+    pages = "249--275",
+    doi = "10.1162/evco_a_00069"
```
pat-s, line 26: @lorenzwalthert DO you have an idea why this alignment is not honoured?

R/mlr_reflections.R
```diff
@@ -74,9 +74,9 @@ local({
   ### Task
   # task types + constructors
   mlr_reflections$task_types = rowwise_table(.key = "type",
-    ~type,     ~package, ~task,         ~learner,         ~prediction,         ~measure,
-    "regr",    "mlr3",   "TaskRegr",    "LearnerRegr",    "PredictionRegr",    "MeasureRegr",
-    "classif", "mlr3",   "TaskClassif", "LearnerClassif", "PredictionClassif", "MeasureClassif"
+    ~type, ~package, ~task, ~learner, ~prediction, ~measure,
+    "regr", "mlr3", "TaskRegr", "LearnerRegr", "PredictionRegr", "MeasureRegr",
+    "classif", "mlr3", "TaskClassif", "LearnerClassif", "PredictionClassif", "MeasureClassif"
```
pat-s, line 79: @lorenzwalthert DO you have an idea why this alignment is not honoured?

R/LearnerClassifDebug.R
pat-s: 👍 

resetted and restyled - some more alignment takes it seems :)

R/TaskClassif_penguins.R
lorenzwalthert: Because spacing is only honoured one level deep. This is two level deep (function call in function call). Sorry, styler can't do this. Use `strict = FALSE` if you want to preserve one more more space here (but it comes with other implications as shown in the dedicated vignette).

R/bibentries.R
lorenzwalthert: Because title is not aligned xD. You must add some spaces after `title` before `=`.

R/mlr_reflections.R
lorenzwalthert: Because left alignment only works with `=`, not with `~`. See https://github.com/lorenzwalthert/mlr3/pull/2#discussion_r619304282

lorenzwalthert (changes requested):

R/bibentries.R
pat-s: Ah yeah! I only looked at the green part of the diff I guess.

R/mlr_reflections.R
lorenzwalthert: I tried for a few hours and I did not get to work it easily... Maybe I can look into it again since we are on it.

pat-s: This is really niche so really think about if you want to do this :)

Would be great and appreciated of course though 👍

lorenzwalthert: Ok, I found a way to do it and merged it into master: https://github.com/r-lib/styler/pull/785. After the other alignment things were implemented properly, it seemed much easier than I thought initially. Are you happy now? xD

pat-s: Thanks! Just restyled but it seems it not yet working 🙈

lorenzwalthert: Because you have an extra wurst with `.key = '...'` -.-. Fix on the way...

lorenzwalthert: Please try again xD

pat-s:
@lorenzwalthert The latest diff looks good to me! Thanks a lot for your help and persistence here 🙏 

@mllg Are you fine with the diff and hence the style guide? If so, we can merge and I would update the documentation and inform all devs.

R/mlr_reflections.R
lorenzwalthert: This seems resolved with latest styling.

lorenzwalthert:
What’s failing here @pat-s ?

pat-s:
Some git checkout issues, seems random. I think we can ignore it.

Styling is fine 👍

lorenzwalthert:
I think it was a GitHub outage. And now we have a merge conflict. I suggest to resolve it, retrigger styling and then get this merged before more merge conflicts accumulate.

pat-s:
I'd like to have (at least) an approval by @mllg

mllg:
Looks good to me. Thanks @pat-s and @lorenzwalthert for all the work you've put into this.

PR #630, opened 2021-04-16
fix predict newdata

jakob-r:
What failed?
- `$predict_newdata` without y column for a learner that does not support missings

https://github.com/mlr-org/mlr3/blob/893611d497b2299765ccc338d8f16f6108caf85d/tests/testthat/test_Learner.R#L134-L139

Why?
- We called `assert_learnable()` (questionable name here) when calling `$predict` which checked ALL columns if there are missings.

Note: Another fix would be to just change `miss = task$missings()` to `miss = task$missings(cols = task$feature_names)` but then we still have the weird naming of `assert_learnable()` for prediction. Also we probably never ever want to allow missing y in `train()`?

### DIFF
--- a/R/Learner.R
+++ b/R/Learner.R
@@ -236,7 +236,7 @@ Learner = R6Class("Learner",
     #' @return [Prediction].
     predict = function(task, row_ids = NULL) {
       task = assert_task(as_task(task))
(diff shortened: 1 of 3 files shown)


mllg:
NB: Only the github version was buggy.

PR #648, opened 2021-05-27
Add get_base_learner() to ensure extractors work on nested objects

mllg:
Addresses https://stackoverflow.com/questions/67381311/variable-importance-by-mlr3filters-does-not-work-in-mlr3proba-after-preproce.

Todo: 
* [ ] Add S3 method in mlr3tuning for AutoTuner
* [ ] Add S3 method in mlr3pipelines for GraphLearner
* [ ] Add S3 method in mlr3fselect for AutoFSelector


mllg:
@be-marc @mb706 Some extractors like the importance filter do not work with GraphLearner/AutoTuner/AutoFSelector. I suggest they call `get_base_learner(learner)$importance()` instead of `learner$importance()` to fix this. `get_base_learner()` extracts the learner at the end of a graph (and returns NULL if this is not feasible) or the finally fitted learner for AutoTuner/AutoFSelector.

Do you see any problems with this approach? Do you have an idea for a better name?

be-marc:
I don't see a problem for AutoTuner/AutoFSelector.  Nice improvement.

mllg:
Switched from S3 to class method.

Issue #651, opened 2021-05-29
Support for custom, factor-based resampling


pat-s:
As discussed in Mattermost.

Something along the lines of

```
rsmp("custom", split = <factor>)
```

where `<factor>` is a factor vector or a variable name from the task.

mllg:
Will https://github.com/mlr-org/mlr3/blob/main/R/ResamplingCustomCV.R suffice?

mllg:
Solved via #654.

PR #654, opened 2021-06-12
Update `ResamplingCustomCV` and add tests

pat-s:
#651 

@mllg I've found `custom_cv$instantiate(task, split = split_f)` nicer to read than `custom_cv$instantiate(task, f = split_f)`. Feel free to change back if you disagree.

`ResamplingCustomCV` now accepts either an external factor or a feature from the task.

### DIFF
--- a/R/ResamplingCustomCV.R
+++ b/R/ResamplingCustomCV.R
@@ -26,8 +26,8 @@
 #'
 #' # Instantiate Resampling:
 #' custom_cv = rsmp("custom_cv")
-#' split_f = factor(c(rep(letters[1:3], each = 3), NA))
-#' custom_cv$instantiate(task, split = split_f)
+#' f = factor(c(rep(letters[1:3], each = 3), NA))
+#' custom_cv$instantiate(task, f = split_f)
 #' custom_cv$iters # 3 folds
 #'
 #' # Individual sets:
--- a/R/ResamplingCustomCV.R
+++ b/R/ResamplingCustomCV.R
@@ -66,7 +66,7 @@ ResamplingCustomCV = R6Class("ResamplingCustomCV", inherit = Resampling,
         type = NULL
         assert_character(split, len = 1)
         assert_subset(split, task$feature_types[`type` == "factor" | `type` == "character", ]$id)
-        split = as.factor(task$data()[[split]])
+        split = as.factor(task$data(cols = split))
       }
       self$instance = split(task$row_ids, split, drop = TRUE)
       self$task_hash = task$hash
--- a/R/ResamplingCustomCV.R
+++ b/R/ResamplingCustomCV.R
@@ -4,15 +4,14 @@
 #' @include Resampling.R
 #'
 #' @description
-#' Splits data into training and test sets in a cross-validation fashion.
-#' Splits are defined by argument `split` provided during instantiation.
+#' Splits data into training and test sets in a cross-validation fashion based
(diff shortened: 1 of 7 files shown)


R/ResamplingCustomCV.R
```diff
@@ -46,14 +46,21 @@ ResamplingCustomCV = R6Class("ResamplingCustomCV", inherit = Resampling,
...
     #'   Empty factor levels are dropped and row ids corresponding to missing values are removed,
     #'   c.f. [split()].
-    instantiate = function(task, f) {
+    instantiate = function(task, split) {
       task = assert_task(as_task(task))
-      assert_factor(f, len = task$nrow, all.missing = FALSE)
-      self$instance = split(task$row_ids, f, drop = TRUE)
+      if (is.factor(split)) {
+        assert_factor(split, empty.levels.ok = FALSE, len = task$nrow, all.missing = FALSE)
+      } else if (is.character(split)) {
```
mllg, line 57: This is undocumented and not clear to me. If this is a character, you want to split on the columns referenced by this vector? The code in the block does not seem to work if there are multiple columns provided. I also wonder if it would be safer to provide two separate arguments here (mutually exclusive).

R/ResamplingCustomCV.R
```diff
@@ -46,14 +46,21 @@ ResamplingCustomCV = R6Class("ResamplingCustomCV", inherit = Resampling,
...
+        # suppress "no visible binding for global variable" note
+        type = NULL
+        assert_subset(split, task$feature_types[`type` == "factor" | `type` == "character", ]$id)
+        split = as.factor(task$data()[[split]])
```
mllg, line 61: This queries the complete data, but only column `split` seems to be required, thus `task$data(cols = split)` is more efficient.

pat-s: Thanks! Incorporated.

R/ResamplingCustomCV.R
pat-s: > If this is a character, you want to split on the columns referenced by this vector? The code in the block does not seem to work if there are multiple columns provided. 

Yes, when `split` is chr I would like to split by this existing vector within the task.

> I also wonder if it would be safer to provide two separate arguments here (mutually exclusive).

No strong opinion on that, would make documentation more distinct but also might raise confusion which argument should be used for which effect.

I've documented it and added another assertion. If this still feels wrong to you, feel free to change it to your liking :)

mllg:
While working on this, two other bugs emerged and I kind of messed up this PR. Sorry.

I've decided it is safer to have to separate arguments to provide the split variable. Will merge soon and I'll release as soon as mlr3pipelines is uploaded.

pat-s:
Thanks. 

Could using a single character as an argument be problematic in some way? Also in terms of autocompletion? I thought about it and changed it to `split` on purpose. No strong opinion, just wondering since I don't think we have this principle in any other function?

mllg:
> Could using a single character as an argument be problematic in some way? Also in terms of autocompletion? I thought about it and changed it to `split` on purpose. No strong opinion, just wondering since I don't think we have this principle in any other function?

Generally speaking, changing behavior depending on the type is often error-prone because R auto-converts somewhat unexpectedly. For example, if you have a task with a single row, the execution solely depends on the parameter being a character or factor, and this often gets mixed up by users. Having two arguments is not really a drawback AFAICT.

PR #662, opened 2021-07-08
Update run_paramtest to error on extra parameters

RaphaelS1:
This will likely break all downstream packages but arguably they should already be erroring because having too many parameters will confuse users


mllg:
Thanks. I will merge this after the next release.

PR #692, opened 2021-09-26
Force immediate log output in `future_mapply()` for `benchmark()` calls

pat-s:
Origin: https://stackoverflow.com/questions/69282207/mlr3-benchmarking-status-messages-are-only-displayed-after-full-benchmark-is?noredirect=1#comment122482373_69282207

When calling `benchmark()`, currently log output is hold back until the function has finished. This small patch should solve this.

There is a workaround if one sets `options("mlr3.debug" = TRUE)` because this uses `base::mapply()` internally. However, this is not the default (and also never will be).

```r
library("mlr3")
tasks = list(tsk("penguins"), tsk("sonar"))
learners = list(lrn("classif.featureless"), lrn("classif.rpart"))
resamplings = list(rsmp("cv"), rsmp("subsampling"))

grid = benchmark_grid(tasks, learners, resamplings)
print(grid)

# workaround
# options("mlr3.debug" = TRUE)

benchmark(grid)
```

I don't see any downsides currently even though the argument description of `future.stdout` reads as follows:

> If TRUE (default), then the standard output of the underlying futures is captured, and re-outputted as soon as possible. If FALSE, any output is silenced (by sinking it to the null device as it is outputted). If NA (not recommended), output is not intercepted.


mllg:
Setting `future.stdout` to NA will not work for futures running in a different session, e.g. with plan multisession.

I did some testing, and I observed that the delayed output only seems to be an issue for the "sequential" plan? We could set `future.stdout` to `NA` only for sequential execution. Would that help?

pat-s:
Sounds good, I haven't tested parallel execution. 

Thanks for checking!

PR #739, opened 2021-12-03
pkgdown v2 fixes

pat-s:


pat-s:
The svg was preferred by pkgdown over the png in pkgdown v2 (which is good in general) but the svg seems borked.
Hence I deleted it - AFAIK it is not used anywhere.

PR #759, opened 2022-01-19
remove obsolete # nolint

MichaelChirico:
Or at least, it looks like this is obsolete (since the assignment_linter is dropped in mlr3's .lintr config)


pat-s:
Thanks!

PR #777, opened 2022-02-15
refactor: Rename test set

mllg:


sebffischer:
How far done is that branch @mllg?

Issue #770, opened 2022-02-11
Rename row role test -> validation


mllg:
- [ ] Rename row role "validation" to "holdout"
- [ ] Allow row role "validation" as alternative name for "test"
- [ ] Adjust all packages
- [ ] Soft-deprecate "test" with a warning
- [ ] Deprecate "test"

PR #795, opened 2022-04-03
Valid

sebffischer:
This would allow proper support for early stopping for boosting and deep learning


sebffischer:
Maybe this change should come with renaming the test set to validation set, i.e. set 2 here https://github.com/mlr-org/mlr3/issues/770

sebffischer:
I think this PR is necessary because we clearly state that the ids that are stored in the col_role `holdout` are not used for anything. 
Look at test_validation.R to see how I intended how this works 
I think it makes sense that this is part of the train method because conceptually I would say it is not a property of the task but of the train method

PR #798, opened 2022-04-06
feat: Add method $set() to Learner.

sebffischer:
This makes the setting of parameters more convenient than
it used to be.

Example: 

``` r
library("mlr3")
l = lrn("regr.rpart")
l$param_set$values
#> $xval
#> [1] 0
l$set(xval = 2, minsplit = 3)
l$param_set$values
#> $xval
#> [1] 2
#> 
#> $minsplit
#> [1] 3
```

<sup>Created on 2022-04-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/Learner.R
+++ b/R/Learner.R
@@ -351,6 +351,21 @@ Learner = R6Class("Learner",
     },
 
     #' @description
+    #' Sets parameters of the learner. Existing parameters are overwritten if their values are
+    #' changed. Other parameters are left as is.
+    #' @return
+    #' Returns the object itself, but motified **by reference**.
+    set = function(...) {
+      new_pars = list(...)
+      assert_named(new_pars)
+      old_pars = self$param_set$values
+      old_pars = old_pars[names(old_pars) %nin% names(new_pars)]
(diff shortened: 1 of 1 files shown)


mllg:
1. I think we should call `insert_named()` here.
2. Should the name maybe more specific? `set_params()`?
3. We should also introduce a second argument for programming, e.g. `set(..., .values = list())`, to avoid having to use `do.call()` all the time.
4. It would be confusing to have two different objects to set hyperpars: `learner$set()` to add parameters to the existing ones, and `learner$param_set$values()` to overwrite with a set. If we want a convenience function for this, we should use it exclusively, i.e. we would need an additional argument like `insert` to toggle the behaviour, defaulting to `TRUE`.
5. We have many more objects with hyperpars. We would need such a method for all these objects. The helper function to set hyperpars (`set_pars(.ps, ..., .values = list(), .insert = TRUE)`) should go into `paradox`.

mllg:
We agreed to put this helper into paradox ... @be-marc will create a PR.

PR #799, opened 2022-04-06
Tags reflections

sebffischer:
1. Add `param_tags` to reflections
2. Modify autotest for learner to check for those tags.


mllg:
Is there a reason for being particularly restrictive here?

sebffischer:
In mlr3extralearners there were all sorts of tags that I just cleaned up (because you suggested I do so)

sebffischer:
I thought it was the policy to not have arbitrary tags

PR #819, opened 2022-05-12
BREAKING CHANGE: Introduce default_params

sebffischer:
* Improves transparency: Learners should set those values in the
  initialize function (and not e.g. in $.train)
* Could be used in description for automatic display of defaults
* Could be used to properly reset a learner to it's initial state.


sebffischer:
@mllg What do you think?

mllg:
I like the idea. Would this break mlr3learners and other packages? If yes, can we prepare an intermediate release so that we can migrate gracefully, i.e. without synchronizing CRAN package uploads?

sebffischer:
So I think now it should not break anything, and we don't have to change the learners either

PR #820, opened 2022-05-16
fix: ability to print objects wo label 

sebffischer:
'label'


sebffischer:
Otherwise loading older mlr3 learners results in an error when printing

PR #830, opened 2022-06-15
fix: rd_info fails because of roxygen2 7.2

sebffischer:


sebffischer:
I also could not find anything in the news, but it seems to do `paste(..., collapse = "," )` which breaks our tables etc.  
I will raise an issue in roxygen2, but the current solution should work always, no?

sebffischer:
https://github.com/r-lib/roxygen2/issues/1361

PR #831, opened 2022-06-16
refactor: mlr_reflections and TaskUnsupervised

be-marc:
## Reflections and Assertions

* A learner type can now work with different task types e.g.  `LearnerClassif` with `TaskClassif` and `TaskClassifST`
* Assertions are performed on classes instead of `task_type` (`mlr_reflections$task_generators`)
* `Learner$task_type` is replaced by `Learner$learner_type`
* `Measure$task_type` is replaced by `Measure$measure_type`

## Unsupervised Task

* `as_task_unsupervised` converters
* A `Prediction` object from a `TaskUnsupervised` gets its class from the training task e.g. training with `TaskClassif` and `LearnClassif`. Predicting with the learner and `TaskUnsupervised` results in a `PredictionClassif`.

Includes #829


mllg:
Ping me when this is ready. Also, how many packages will this break? :)

be-marc:
We need another approach.

PR #873, opened 2022-10-25
Extend task printer to customize col role printing

### LINKED ISSUE #868: Task printer should print additional col roles
This came up in the context of fairness where we might want to denote the protected attribute. 

One idea would be to simply print all other non-empty `col_roles`.
If this become to much / too bloated we can also add reflections but hopefully we can do without?

mllg:
Fixes #868 .

### DIFF
--- a/R/Task.R
+++ b/R/Task.R
@@ -174,7 +174,7 @@ Task = R6Class("Task",
 
       # print additional columns are specified in reflections
       before = mlr_reflections$task_print_col_roles$before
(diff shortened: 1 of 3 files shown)


mllg:
@pfistfl this is the best I came up with. You would have to add "pta" to the new reflection, i.e.

```r
mlr_reflections$task_print_col_roles$after = union(mlr_reflections$task_print_col_roles$after, c("PTA" = "pta"))
```

Issue #816, opened 2022-05-06
oob_error measure breaks for survival predictions


RaphaelS1:
See https://github.com/mlr-org/mlr3proba/issues/275

Reprex:

``` r
library(mlr3proba)
library(mlr3learners)

l = lrn("surv.ranger")
t = tsk("rats")
l$train(t)
p = l$predict(t)
m = msr("oob_error")

p$score(m, learner = l)
#> oob_error 
#>       NaN
l$oob_error()
#> [1] 0.25694

m$predict_type
#> [1] "response"
p$predict_types
#> [1] "crank" "distr"
```

<sup>Created on 2022-05-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Returning `NaN` because of the different predict types. Not sure if the best solution is for me to implement `surv.oob_error` that has `distr` predict type or something else?

https://github.com/mlr-org/mlr3/blob/da45a5acb7a1f4635f278774488f7c515e1b8380/R/MeasureOOBError.R#L26

bblodfon:
Hi @sebffischer and @RaphaelS1! Has there been a discussion on how to tackle this? It would be nice to not restrict the `predict_type` of the `oob_error` and be able to tune a Random Forest survival learner using an `insample` resampling (train equals to test set) and instead of the overestimated train error performance get the `oob_error` of the learner.

Similar discussion in [mlr](https://github.com/mlr-org/mlr/issues/338).

sebffischer:
My proposal would be the following: 


1. Allow for the Measure OOB to support multiple predict types, i.e. if any one the supported predict types is present, the measure is applicable. This would require to change the following check in `score_single_measure`: 

```r
  if (!is_scalar_na(measure$predict_type) && measure$predict_type %nin% prediction$predict_types) {
    # TODO lgr$debug()
    return(NaN)
  }
```
to 
```r
  if (!is_scalar_na(measure$predict_type) && any(measure$predict_type %nin% prediction$predict_types)) {
    # TODO lgr$debug()
    return(NaN)
  }
```

2. We need to add the possibility to extend the predict types of a Measure when loading a package. 
I.e. I imagine something like

```r
mlr3::MeasureOOBError$set("public", "predict_type", c("response", "distr", "crank"))
```

Note that this currently does not work, as this value is overwritten during the initialization of the measure. 

3. `mlr3proba` needs to add the respective predict types in its `.onLoad` function and needs to remove them again in the `.onUnload` function. 

[...]

PR #889, opened 2023-01-24
fix: change predict_type for generic measures

mllg:
Addressing #816


sebffischer (approved):

mllg:
@RaphaelS1 does this solve #816 ?

RaphaelS1:
Looks like it should. Happy for @sebffischer to test and decide if it makes sense (saw his ideas in the other issues) and approve if so

PR #895, opened 2023-02-03
feat: add print functions to data.table output

be-marc:
Test

```r
design = benchmark_grid(
  tasks = tsk("iris"),
  learners = lrn("classif.rpart"),
  resamplings = rsmp("cv", folds = 3)
)

bmr = benchmark(design)
bmr
bmr$score()
bmr$aggregate()

rr = resample(tsk("iris"), lrn("classif.rpart"), rsmp("cv", folds = 3))
rr
rr$score()

design
```

I think we can't avoid to copy the table once when printing, because the `print.data.table()` function has no `column` argument.


be-marc:
Also fixes bug in `as_result_data()`. Predictions were stored as list of predictions but it should be list of lists of predictions e.g. `list(list(test = Prediction))`.

R/BenchmarkResult.R
```diff
@@ -99,6 +99,7 @@ BenchmarkResult = R6Class("BenchmarkResult",
     #' Printer.
     print = function() {
       tab = self$aggregate(measures = list(), conditions = TRUE)
+      setattr(tab, "class", c("data.table", "data.frame"))
```
mllg, line 102: can you proceed without dropping the class, e.g. by calling `print.data.table()` instead of `print()`?

R/BenchmarkResult.R
```diff
@@ -177,6 +178,8 @@ BenchmarkResult = R6Class("BenchmarkResult",
 
       set(tab, j = "prediction", value = as_predictions(tab$prediction, predict_sets))
 
+      setattr(tab, "class", c("bmr_score", class(tab)))
```
mllg, line 181: this should go into a helper function

R/BenchmarkResult.R
```diff
@@ -500,3 +505,19 @@ c.BenchmarkResult = function(...) { # nolint
   init = BenchmarkResult$new()
   Reduce(function(lhs, rhs) lhs$combine(rhs), bmrs, init = init)
 }
+
+#' @export
+print.bmr_score = function(x, ...) {
+  setattr(x, "class", c("data.table", "data.frame"))
+  print(x[, !c("uhash", "task", "learner", "resampling", "prediction")])
+  cat("Hidden columns: uhash, task, learner, resampling, prediction")
+  setattr(x, "class", c("bmr_aggregate", "data.table", "data.frame"))
```
mllg, line 514: :(

mllg: wrong class in return

R/BenchmarkResult.R
```diff
@@ -500,3 +505,19 @@ c.BenchmarkResult = function(...) { # nolint
...
+}
+
+#' @export
+print.bmr_aggregate = function(x, ...) {
+  setattr(x, "class", c("data.table", "data.frame"))
+  print(x[, !"resample_result"])
+  cat("Hidden columns: resample_result")
+  setattr(x, "class", c("bmr_aggregate", "data.table", "data.frame"))
```
mllg, line 522: :(

R/ResampleResult.R
```diff
@@ -341,3 +344,11 @@ c.ResampleResult = function(...) {
 resample_result_aggregate = function(rr, measures) {
   set_names(map_dbl(measures, function(m) m$aggregate(rr)), ids(measures))
 }
+
+#' @export
+print.rr_score = function(x, ...) {
+  setattr(x, "class", c("data.table", "data.frame"))
+  print(x[, !c("task", "learner", "resampling", "prediction")])
+  cat("Hidden columns: task, learner, resampling, prediction")
+  setattr(x, "class", c("bmr_aggregate", "data.table", "data.frame"))
```
mllg, line 353: wrong class in return

R/ResampleResult.R
```diff
@@ -341,3 +344,11 @@ c.ResampleResult = function(...) {
 resample_result_aggregate = function(rr, measures) {
   set_names(map_dbl(measures, function(m) m$aggregate(rr)), ids(measures))
 }
+
+#' @export
+print.rr_score = function(x, ...) {
+  setattr(x, "class", c("data.table", "data.frame"))
+  print(x[, !c("task", "learner", "resampling", "prediction")])
+  cat("Hidden columns: task, learner, resampling, prediction")
```
mllg, line 352: DRY (don't repeat yourself)

add a helper function to call from your printers.

R/as_result_data.R
```diff
@@ -12,7 +12,7 @@
 #' @param learners (list of trained [Learner]s).
 #' @param resampling ([Resampling]).
 #' @param iterations (`integer()`).
-#' @param predictions (list of [Prediction]s).
+#' @param predictions (list of list of [Prediction]s).
```
mllg, line 15: what's happening here? why do we need to change the internal data structure for a printer?

mllg (changes requested):

R/BenchmarkResult.R
be-marc: Unfortunately, we cannot access `print.data.table()`. At least `settattr()` works by reference.

R/as_result_data.R
be-marc: There was a bug in line 48.  The new printer is called in the example and failed.

be-marc: We don't use the function so we didn't notice the bug.

PR #916, opened 2023-03-30
feat: Resampling saves param vals during instantiation

sebffischer:
Some inference method that e.g. provide confidence intervals for the generalization error need access to these parameter values. Currently we had to trust the parameter values are not changed after instantiation, which is suboptimal.


mllg:
If the only purpose is to protect paramsets from modification after initialization, there are better ways (make `param_set` an active binding).

PR #954, opened 2023-08-24
Vectorize params in benchmark

mllg:


be-marc:
revdepcheck shows new errors in mlr3batchmark and mlr3pipelines. pipelines seems to be unrelated.

PR #979, opened 2023-11-30
fix: reduce number of data table threads when running with future

be-marc:


mllg:
+ https://cran.r-project.org/web/packages/RhpcBLASctl/index.html
+ test for multicore

PR #981, opened 2023-12-07
fix: remove task prototype when resample

be-marc:


R/worker.R
```diff
@@ -262,6 +262,12 @@ workhorse = function(iteration, task, learner, resampling, param_values = NULL,
     learner$state$model = NULL
   }
 
+  # repeated saving of the prototype leads to large ResultData objects if the task contains many columns, factor levels or attributes
+  learner$state$data_prototype = NULL
+  learner$state$task_prototype = NULL
+  learner$fallback$state$data_prototype = NULL
```
sebffischer, line 268: check first whether fallback exists.

sebffischer (changes requested):
Also I think we should: 

* [ ] Add the column names to the learner state, then after the new version of mlr3 is released, we can change the `ordered_features()` function from mlr3learners/ mlr3extralearners
* [ ] Start removing the unused data_prototype (?)

be-marc:
> Start removing the unused data_prototype (?)

I will open an PR at vetiver.

R/worker.R
be-marc: There is no error even if the fallback does not exist.

be-marc:
Blocked by https://github.com/mlr-org/mlr3learners/pull/278.

Issue #891, opened 2023-01-24
[Suggestion] Serialization property for learners


sebffischer:
Some learners (especiall torch learners) but also e.g. lightgbm break when saved and reloaded. 
One situation where this occurs is if one calls `benchmark(..., store_models = TRUE)`. 

To avoid this / make it more comfortable for the user, I suggest adding a property `"serialize"`. 
If this property is present, a learner must implement a public method `serialize()` that converts the learners `$state` into a serialized state. 

To implement that, we could save the previous `$state` of a learner in a private field `$.state` and make `$state` an active binding that unserializes a earner's state if it is accessed and serialized. 

This allows us to hide the serialization from the user in some circumstances e.g. in the `benchmark()` function we can call `learner$serialize()` if `store_models` is `TRUE`. The user can then afterwards access the learner and does not have to call `learner$unserialize()` because this will automatically happen when he accesses the state. 

E.g. using [bundle](https://github.com/rstudio/bundle), this might look as follows for LightGBM: 

```r
LearnerClassifLightGBM = R6Class("LearnerClassifLightGBM",
  ...,
  public = list(
    serialize = function() {
      private$.state = bundle(private.$state)
      private$.serialized = TRUE
    }
  ),
  private = list(
    .state = NULL,
    .serialized = FALSE
  ),
  active = list(
    state = function(x) {
      if (missing(x)) {
        if (private$.serialized) {
          private$.state = unbundle(private$.state)
          private$.serialized = FALSE
        }
      } else {
        private$.state = x
      }
      return(private$.state)
    }
  )
)
```

In addition to that, it might be convenient to offer a `$save(path)` method that calls `$serialize()` and then `saveRDS()`

sebffischer:
Note that there is also an open issue in bundle: https://github.com/rstudio/bundle/issues/13

sebffischer:
There is also this related package: https://github.com/HenrikBengtsson/marshal

sebffischer:
this is implemented

PR #993, opened 2024-01-25
feat(Learner): support marshal property

sebffischer:
Suggestion: 
* Per default, when calling `benchmark()` or `resample()`, the models are in unmarshaled form.
* When calling `$score()` and `$aggregate()` it is checked whether the measure needs access to the model, and if so, a good error message is given that informs the user that:
    a) He needs to `$unmarshal()` the object
    b) After doing so, he needs to `$marshal()` again to be able to save the object

TODOs:
* we can save one cycle when using callr encapsulation + parallelization by not discarding the marshalled model before the predict step. 
* talk about the cloning issue (see https://github.com/HenrikBengtsson/marshal/issues/12)

Points for discussion: 
* package versioning: It would be useful to store the most important package versions when marshaling, as it will otherwise become tricky to unmarshal the object later. Henrik's approach is to store the unmarshal function in the marshaled object: https://github.com/HenrikBengtsson/marshal/blob/72e976c61f00861ceb0b4c0c0fd7016b4eee417a/R/marshal.default.R#L30 and it is then the obligation of the unmarshal function to do the version checking. 

This is useful for learners that e.g. rely on external pointers and that need processing of the model to be serializable.
Addresses https://github.com/mlr-org/mlr3/issues/891

Input checks on the measure (check that model is not marshaleld when accessing the model) only work partially (https://github.com/mlr-org/mlr3/issues/1000)

* mlr3tuning: https://github.com/mlr-org/mlr3tuning/pull/410
* mlr3batchmark: https://github.com/mlr-org/mlr3batchmark/pull/28
* mlr3pipelines https://github.com/mlr-org/mlr3pipelines/pull/759

TODOs:
* [ ] Gallery Post


sebffischer:
Related PR in mlr3torch: 
https://github.com/mlr-org/mlr3torch/pull/179

R/Learner.R
```diff
@@ -164,6 +164,10 @@ Learner = R6Class("Learner",
       self$predict_types = assert_ordered_set(predict_types, names(mlr_reflections$learner_predict_types[[task_type]]),
         empty.ok = FALSE, .var.name = "predict_types")
       private$.predict_type = predict_types[1L]
+
+      if (!is.null(private$.bundle) && !is.null(private$.unbundle)) {
```
be-marc, line 168: I would keep Michel's style and set "bundle" as a property in the initialize of the learner.

R/Learner.R
```diff
@@ -388,6 +428,12 @@ Learner = R6Class("Learner",
       self$state$model
     },
 
+    #' @field bundled (`logical(1)` or `NULL`)\cr
+    #' Indicates whether the model has been bundled (`TRUE`), unbudled (`FALSE`), or neither (`NULL`).
```
be-marc, line 432: Is there a more elegant solution? Somehow strange that you always have to use `isTRUE()` because of the three states.

R/Learner.R
sebffischer: This case is different because we can infer the property because it is something that is implemented in `mlr3` and not in the upstream package. I don't really mind but I don't see the benefit of adding a possible source of error.

sebffischer:
* `$bundle()` and `$unbundle()` should only be implemented in the Learners that actually implement them.

R/Measure.R
```diff
@@ -184,6 +186,9 @@ Measure = R6Class("Measure",
       if ("requires_model" %in% self$properties && (is.null(learner) || is.null(learner$model))) {
         stopf("Measure '%s' requires the trained model", self$id)
       }
+      if ("requires_model" %in% self$properties && marshalled_model(learner$model)) {
+        stopf("Measure '%s' requires the trained model, but model is un marshalled form", self$id)
```
sebffischer, line 190:
```suggestion
        stopf("Measure '%s' requires the trained model, but model is marshalled form", self$id)
```

man-roxygen/param_learner_properties.R
```diff
@@ -7,3 +7,5 @@
 #'   * `"importance"`: The learner supports extraction of importance scores, i.e. comes with an `$importance()` extractor function (see section on optional extractors in [Learner]).
 #'   * `"selected_features"`: The learner supports extraction of the set of selected features, i.e. comes with a `$selected_features()` extractor function (see section on optional extractors in [Learner]).
 #'   * `"oob_error"`: The learner supports extraction of estimated out of bag error, i.e. comes with a `oob_error()` extractor function (see section on optional extractors in [Learner]).
+#'   * `"marshal"`: To save learners with this property, you need to call `$marshal()` first.
+#'      If a learner is in a marshalled state, you call first need to call `$unmarshal()` to use it's model, e.g. for prediction.
```
sebffischer, line 11:
```suggestion
#'      If a learner is in a marshalled state, you call first need to call `$unmarshal()` to use its model, e.g. for prediction.
```

R/Measure.R
```diff
@@ -184,6 +185,9 @@ Measure = R6Class("Measure",
       if ("requires_model" %in% self$properties && (is.null(learner) || is.null(learner$model))) {
         stopf("Measure '%s' requires the trained model", self$id)
       }
+      if ("requires_model" %in% self$properties && marshalled_model(learner$model)) {
+        stopf("Measure '%s' requires the trained model, but model is marshalled form", self$id)
```
sebffischer, line 189:
```suggestion
        stopf("Measure '%s' requires the trained model, but model is in marshalled form", self$id)
```

R/marshal.R
```diff
@@ -0,0 +1,131 @@
...
+}
+
+#' @rdname marshaling
+#' @export
+marshaled_model = function(model) {
+  test_class(model, "marshaled")
+}
+
+#' @export
+marshal_model.default = function(model, ...) {
```
sebffischer, line 99: remove this method as it is never called.

sebffischer: do we really need the 'marshal' property.

sebffischer: is marshaling defined as `unmarshal(marshal(x))` returns `x` or does it also include that `marshaled(marshal(x))` returns `TRUE`.

R/mlr_reflections.R
```diff
@@ -125,7 +125,7 @@ local({
   )
 
   ### Learner
-  tmp = c("featureless", "missings", "weights", "importance", "selected_features", "oob_error", "loglik", "hotstart_forward", "hotstart_backward")
+  tmp = c("featureless", "missings", "weights", "importance", "selected_features", "oob_error", "loglik", "hotstart_forward", "hotstart_backward", "marshal")
```
sebffischer, line 128: do we really need this?

R/LearnerClassifLily.R
```diff
@@ -0,0 +1,74 @@
...
+    marshal = function() {
+      learner_marshal(self)
+    },
+    #' @description
+    #' Unmarshal the learner.
+    unmarshal = function() {
+      learner_unmarshal(self)
+    }
+  ),
+  active = list(
```
sebffischer, line 40: move this learner to `inst`

R/mlr_reflections.R
sebffischer: --> we either keep the property or the default marshal method

sebffischer: Even if we identify this I think it is nice to have this in the overview of the learner properties

PR #996, opened 2024-01-30
feat(Learner): Allow for additional input checks on task

sebffischer:
By adding this feature, it is possible for learner's to perform custom compatibility checks. This is e.g. needed in torch, where we have a "tensor column", where the tensor can have different shapes. To verify that the tensor shapes fits together with the neural network, we want to perform input checks **before** calling into train, as incompatability is a user error and should not trigger the fallback learner

* [ ] In case this is accepted, add it do the "Extending" tutorial from mlr3extralearners.

### DIFF
--- a/R/Learner.R
+++ b/R/Learner.R
@@ -85,7 +85,7 @@
 #' @section Additional Task Checks:
 #' Learner may perform custom compatibility checks on a task that determine whether a learner is applicable to a task
 #' using the optional `$check_task(task)` public method.
-#' When providing this method, it should either return `TRUE` or return the an error message as a `character(1)`.
+#' When providing this method, it should either return `TRUE` or return an error message as a `character(1)`.
 #'
 #' @template seealso_learner
 #' @export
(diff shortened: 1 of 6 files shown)


sebffischer:
* [ ] In case this is accepted, add it do the "Extending" tutorial from mlr3extralearners.

R/Learner.R
```diff
@@ -83,6 +82,11 @@
 #' lrn$param_set$add(paradox::ParamFct$new("foo", levels = c("a", "b")))
 #' ```
 #'
+#' @section Additional Task Checks:
+#' Learner may perform custom compatibility checks on a task that determine whether a learner is applicable to a task
+#' using the optional `$check_task(task)` public method.
+#' When providing this method, it should either return `TRUE` or return the an error message as a `character(1)`.
```
sebffischer, line 88:
```suggestion
#' When providing this method, it should either return `TRUE` or return an error message as a `character(1)`.
```

sebffischer:
We can alternatively allow for something like this, by introducing a error condition such as `NoFallback` that is then handled by `learner_train()`

PR #1001, opened 2024-02-02
tests: encapsulation method use the same rng state

be-marc:


R/LearnerClassifDebug.R
```diff
@@ -105,7 +105,7 @@ LearnerClassifDebug = R6Class("LearnerClassifDebug", inherit = LearnerClassif,
       }
 
       model = list(response = as.character(sample(task$truth(), 1L)), pid = Sys.getpid(), iter = pv$iter,
-        id = UUIDgenerate())
+        id = UUIDgenerate(), random_number = sample(seq(100000), 1))
```
mb706, line 108: You can just do `sample(100000, 1)` to get the same result

PR #1039, opened 2024-06-21
refactor: use `deparse1(substitute(x))` instead of `deparse(substitute(x))` where a string is expected

m-muecke:


be-marc:
Thanks!

PR #1047, opened 2024-07-08
Feat/loss funs

sebffischer:
Related work: 

* PR in mlr3measures: https://github.com/mlr-org/mlr3measures/pull/45
* the `mlr3inference` package: https://github.com/mlr-org/mlr3inference

Summary of changes:
* `Measure`s can now return more than one scalar value. This is needed to make CIs work, which return a named vector with three elements (estimate, lower, upper)
* Some measures like RMSE, use `$obs_loss` + trafo. In order to be able to create CIs for such measures, this PR adds the `$trafo` field, which in the case of RMSE is `sqrt()`
* Some resampling methods in mlr3inference do not use all resampling iterations for point estimates (e.g. some are only used for variance estimation). We want to exclude these iterations when doing `rr$aggrgate(msr("classif.acc"))`. Therefore, the `Resampling` class can now implement `$primary_iters`, which should be those iterations that should be used to obtain a point estimate. However, some measures with `average = "custom"` cannot handle this. Therefore I added the new `"primary_iters"` property to Measures, which should by default NOT be present in measures that have `average = "custom"`.


R/ResampleResult.R
```diff
@@ -357,7 +372,12 @@ c.ResampleResult = function(...) {
 
 
 resample_result_aggregate = function(rr, measures) {
-  set_names(map_dbl(measures, function(m) m$aggregate(rr)), ids(measures))
+  scores = unlist(map(unname(measures), function(m) {
```
sebffischer, line 375:
```suggestion
  unlist(map(unname(measures), function(m) {
```

Issue #1016, opened 2024-04-11
Column role "weight"


sebffischer:
Currently it only says "Observation weights. Not more than one numeric column may have this role." in the Task documentation. 
Also, I got the reasonable question whether these are also applied during resampling.

sebffischer:
There is a very old and open PR regarding weights for resampling: https://github.com/mlr-org/mlr3/pull/239

sebffischer:
Also it was asked whether the weights are applied when scoring predictions via measures.

berndbischl:
> Currently it only says "Observation weights. Not more than one numeric column may have this role." in the Task documentation.

is there a question related to this?

berndbischl:
* we will improve docs to make clear what currently happens. what currently happens is that these weights are handed down to each learner, in training, when the learner supports a weighted fit. if the learners doesn't support weights, they are ignored

berndbischl:
* we can also have the weights being used in measures

```
task$set_col_role("my_weights", "weights")
my_learner = lrn("regr.rpart", use_weights = TRUE) # ???
my_resa = rsmp("cv")
rr = resample(task, my_learner, my_resa)
mm = msr("regr.mse", task_weights = TRUE)
rr$score(mm)
```
this control where weights are used

berndbischl:
in task: col_roles: weights_train, weights_measures
maybe: weights_resampling

berndbischl:
we also want flags in learner and measure, to control wether the weights are used in computation. also we need to ensure that the docs are clear. maybe the book

berndbischl:
closed by PR #1065

PR #1065, opened 2024-08-11
refactor: weights

mllg:
Working on #1016 and #239.


berndbischl:
why dont we add "use.weights" in the abstract base class of "Learner" if the property "weights" is in the learner?

this not only saves extra code -- in the current system we can now also be inconsistent, so the property can be there but the param not -- or vice versa

berndbischl:
' The weights do not necessarily need to sum up to 1, they are normalized by dividing by the sum of weights.

--> this is now inconsistent with the learner. i would remove that? maybe? doesn't this also depend on aggregation, whether the measure is summed, or not?

PR #1076, opened 2024-08-16
feat: partition function creates training test and validation sets

be-marc:


bblodfon:
@be-marc no stratification anymore? do I need to update the [mlr3proba](https://github.com/mlr-org/mlr3proba/blob/main/R/partition.R) `partition`?

PR #1086, opened 2024-08-17
feat: allow quantile predictions for regression

mllg:
Missing:

- [x] Discussion about names / API, double check if we want to use `quantile` (singular) or `quantiles` (plural) and if this is consistent
- [ ] Add autotest 
- [x] Test with a real learner


be-marc:
https://github.com/mlr-org/mlr3learners/pull/308

Issue #886, opened 2023-01-17
run_autotest should be able to generate parameters depending on the tasks


sebffischer:
Some learners require (such as priority lass from mlr3extralearners) require to set a parameter to a specific value, that depends on properties of the tasks (in this case, the number of features). 

Currently the `run_autotest` function does not really support this. 

Suggestion: add a `parfn` argument that takes in arguments `task, learner` and outputs a list with the new parameters.

PR #1096, opened 2024-08-18
feat(tests): allow to configure learner depending on the task

sebffischer:
Fixes https://github.com/mlr-org/mlr3/issues/886


mllg:
If you want to autotest tasks with `internal_valid`, you can also just create one via `generate_tasks()` after querying the properties of the learner.

sebffischer:
This is for something else. Sometimes learners don't run on a task unless specific task-dependent parameters are set. Currently there is no way to configure something like this.

berndbischl:
this is not even exposed API, "just" unit test code and it solves a clear problem. 
merge now.

PR #1097, opened 2024-08-18
tests: remove unneded marshaling test

### LINKED ISSUE #1083: remove some unneeded marshal tests
they currently slow down the tests

sebffischer:
This functionality is essentially the same as for resample() where there are also the same tests, so this can be removed

Fixes #1083


sebffischer:
@mllg if this is still not good enough I suggest adding a `skip_on_cran()` to the tests so they can also be skipped locally if required.

mllg:
I don't think there is a "good enough".

Please just double check that you don't perform the same tests multiple times and also check if you can make them less expensive. CRAN is currently not the problem, but we should keep the timings reasonable to not end up like mlr2.

Issue #1094, opened 2024-08-18
Feature Request: Allow no required predict sets for a `Measure`


sebffischer:
Some mlr3 Measures don't require a prediction object to operate on, such as `msr("selected_features")` or `msr("internal_valid_score")`. 

Let's say I want to resample xgboost with early stopping and I only care about the internal validation scores (e.g. in the context of tuning). I could then avoid making predictions on the test set alltogether. What I **want** to do is therefore:

```r
l = lrn("classif.xgboost", nrounds = 100,
  early_stopping_rounds = 10, validate = "test", predict_sets = NULL)
  
rr = resample(tsk("iris"), l, rsmp("holdout")

rr$aggregate(msr("internal_valid_score"))
```

The reason why the code above does not run is because each measure must have at least one required predict sets because of this check: https://github.com/mlr-org/mlr3/blob/3f159d91735b2b229605f34f35f22765749dd0c9/R/Measure.R#L147

I.e., for `MeasureInternalValidScore` I need to pick `"test"`, `"train"` or `"internal_valid"`, even though none are required.
The last call to `$aggregate()` will therefore err, no matter how I define the `MeasureInternalValidScore`.

mllg:
You can toggle `empty.ok` and check if the unit tests still pass. But to be fair, not wanting to look at predictions at all is an unusual setting ;)

PR #1100, opened 2024-08-20
feat(measure): allow empty predict_sets

sebffischer:
Solves https://github.com/mlr-org/mlr3/issues/1094

Changes since review:
* I now added the property `"no_prediction_needed"` because survival measures can access both the model and the prediction.
* I changed the defaults of `predict_sets` to `NULL` for those that don't need it. It previously was `"test"`.


berndbischl:
> You can toggle empty.ok and check if the unit tests still pass. But to be fair, not wanting to look at predictions at all is an unusual setting ;)

this is not completely unusual.... this hold for every metric which only refers to the model, and we already have quite a few of those. and if we merge this in, which should probably make a list. 

but things like train_time, measures like AIC, and so on are in here.

R/Measure.R
```diff
@@ -198,8 +195,7 @@ Measure = R6Class("Measure",
     #' @return `numeric(1)`.
     score = function(prediction, task = NULL, learner = NULL, train_set = NULL) {
       assert_measure(self, task = task, learner = learner)
-      assert_prediction(prediction)
-
+      assert_prediction(prediction, null.ok = "requires_model" %nin% self$properties)
```
sebffischer, line 198:
```suggestion
      assert_prediction(prediction, null.ok = "requires_no_prediction" %nin% self$properties)
```

PR #1119, opened 2024-08-24
BREAKING CHANGE: loglik, AIC, BIC, feature-creep

berndbischl:
WARNING: BEFORE WE MERGE:

aic and bic should be removed from book, loglik too
check mlr3learners
mlr3tuning autotuner
add to NEWS


mllg:
Also needs to be removed from mlr3learners for log reg, multinom and linear regression. Or could just stay in the package with improved documentation.

mllg:
TBH, I'd really like to keep AIC/BIC in the packages. I show this during teaching together with a SFS and compare the it to `step()`.

mllg:
After a closer look, we could remove the properties and methods altogether but keep the measures for AIC and BIC. The user then has to ensure that `stats::logLik()` is properly working on the fitted model.

R/Resampling.R
```diff
@@ -46,6 +46,14 @@
 #' Next, the grouping information is replaced with the respective row ids to generate training and test sets.
 #' The sets can be accessed via `$train_set(i)` and `$test_set(i)`, respectively.
 #'
+#' @section Weights:
+#'
+#' Many resamlings support observation weights, indicated by their property `"weights"`.
+#' The weights are stored in the [Task] where the column role `weights_resampling` needs to be assigned to a single numeric column.
+#' The weights are automatically used if found in the task, this can be disabled by setting the hyperparamerter `use_weights` to `FALSE`.
+#' If the resampling is set-up to use weights but the task does not have a designated weight column, an unweighted version is calculated instead.
+#' The weights do not necessarily need to sum up to 1, they are passed down to argument `prob` of [sample()].
```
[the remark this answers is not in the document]
sebffischer: I think here it should also be mentioned what is mentioned in the parameter description below, namely that the weights only influence how the training data is sampled. Maybe we can also mention here that weighting of test observations when calculating performance is possible via the `weights_measure` property.

R/Resampling.R
```diff
@@ -168,19 +180,24 @@ Resampling = R6Class("Resampling",
       task = assert_task(as_task(task))
       strata = task$strata
       groups = task$groups
+      weights = task$weights_resampling$weight
+
+      use_weights = "weights" %in% self$properties && !isFALSE(self$param_set$values$use_weights) && !is.null(weights)
+
+      if (sum(!is.null(strata), !is.null(groups), use_weights) >= 2L) {
+        stopf("Stratification, grouping and weighted resampling are mutually exclusive")
```
sebffischer, line 188: Why do we have this restriction? At least for groups we could sum up the weights and also for strata we could also use the weights within each stratum?

R/Resampling.R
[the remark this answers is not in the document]
sebffischer: Maybe also mention that this still does not necessarily mean that observations with a large weight can appear twice in the train set.

R/ResamplingHoldout.R
```diff
@@ -59,10 +61,11 @@ ResamplingHoldout = R6Class("ResamplingHoldout", inherit = Resampling,
     }
   ),
   private = list(
-    .sample = function(ids, ...) {
+    .sample = function(ids, task, weights, ...) {
+      pv = self$param_set$values
       n = length(ids)
       in_train = logical(n)
-      in_train[sample.int(n, round(n * self$param_set$values$ratio))] = TRUE
+      in_train[sample.int(n, round(n * self$param_set$values$ratio), prob = weights)] = TRUE
```
sebffischer, line 68: Is this here really the semantics that we want?
A task with 10 obs, ratio = 0.8 and a probability vector `c(0, 0, 0, 0, 0, 1, 1, 1, 1, 1)` will not produce a valid resampling split. I am not sure how intuitive this weight semantic is for resamplings that don't allow for duplicate observations in the training data.

R/Task.R
```diff
@@ -939,19 +945,28 @@ Task = R6Class("Task",
     #'   For each resampling iteration, observations of the same group will be exclusively assigned to be either in the training set or in the test set.
     #'   Not more than a single column can be associated with this role.
     #' * `"stratum"`: Stratification variables. Multiple discrete columns may have this role.
-    #' * `"weight"`: Observation weights. Not more than one numeric column may have this role.
+    #' * `"weights_learnering"`: If the task has observation weights with this role, they are passed to the [Learner] during train.
```
sebffischer, line 948:
```suggestion
    #' * `"weights_learner"`: If the task has observation weights with this role, they are passed to the [Learner] during train.
```

R/Task.R
```diff
@@ -939,19 +945,28 @@ Task = R6Class("Task",
...
+    #'    The use of weights can be disabled via by setting the measure's hyperparameter `use_weights` to `FALSE`.
+    #' * `"weights_resampling"`: If the task has observation weights with this role, they are passed to the [Resampling] for weighted sampling.
+    #'    The weights are only used if the resampling's hyperparameter `use_weights` is set to `TRUE`.
     #'
     #' `col_roles` is a named list whose elements are named by column role and each element is a `character()` vector of column names.
     #' To alter the roles, just modify the list, e.g. with \R's set functions ([intersect()], [setdiff()], [union()], \ldots).
     #' The method `$set_col_roles` provides a convenient alternative to assign columns to roles.
+    #'
+    #' The roles `weights_learner`, `weights_measure` and `weights_resampling` may only point to a single numeric column, but they can
+    #' all point to the same column or different columns. Weights must be non-negative numerics with at least one weight being > 0.
```
sebffischer, line 960: weights_resampling usually require more than one weight being > 0

R/Task.R
```diff
@@ -1056,23 +1071,69 @@ Task = R6Class("Task",
...
-    weights = function(rhs) {
+    #' Returns `NULL` if there are is no column with the designated role.
+    weights_learner = function(rhs) {
+      assert_has_backend(self)
+      assert_ro_binding(rhs)
+      weight_cols = private$.col_roles[["weights_learner"]]
+      if (length(weight_cols) == 0L) {
+        return(NULL)
+      }
+      data = self$backend$data(private$.row_roles$use, c(self$backend$primary_key, weight_cols))
```
sebffischer, line 1096: This might fail because columns can be returned in arbitrary order: https://github.com/mlr-org/mlr3/blob/57b6109355ae5219d735a89c188d1b987ae2ff41/R/DataBackendDataTable.R#L61

Instead you should use the `$data()` method of the task: https://github.com/mlr-org/mlr3/blob/57b6109355ae5219d735a89c188d1b987ae2ff41/R/Task.R#L265

R/Task.R
```diff
@@ -1056,23 +1071,69 @@ Task = R6Class("Task",
...
+    #' Returns `NULL` if there are is no column with the designated role.
+    weights_measure = function(rhs) {
       assert_has_backend(self)
       assert_ro_binding(rhs)
-      weight_cols = private$.col_roles$weight
+      weight_cols = private$.col_roles[["weights_measure"]]
       if (length(weight_cols) == 0L) {
         return(NULL)
       }
       data = self$backend$data(private$.row_roles$use, c(self$backend$primary_key, weight_cols))
```
sebffischer, line 1115: same as above

R/Task.R
```diff
@@ -1189,12 +1250,23 @@ task_set_roles = function(li, elements, roles = NULL, add_to = NULL, remove_from
 }
 
 task_check_col_roles = function(self, new_roles) {
-  for (role in c("group", "weight", "name")) {
+  if ("weight" %in% names(new_roles)) {
+    stopf("Task role 'weight' is deprecated, use 'weights_learner' instead")
```
sebffischer, line 1254: If we throw an error here I don't think "deprecated" is the right word? It is already not supported anymore?

inst/testthat/helper_autotest.R
```diff
@@ -233,6 +233,19 @@ run_experiment = function(task, learner, seed = NULL, configure_learner = NULL)
   learner$encapsulate = c(train = "evaluate", predict = "evaluate")
 
   stage = "train()"
+
+  # enable weights
+  if ("weights" %in% learner$properties) {
+    # learner must have flag "use_weights"
+    msg = checkmate::check_subset("use_weights", learner$param_set$ids(), empty.ok = FALSE)
```
sebffischer, line 240:
```suggestion
    msg = checkmate::check_subset("use_weights", learner$param_set$ids())
```

R/Learner.R
```diff
@@ -566,7 +575,10 @@ Learner = R6Class("Learner",
       private$.encapsulate = insert_named(default, rhs)
 
       if (is.null(private$.fallback)) {
-        self$fallback = lrn(mlr_reflections$learner_fallback[[self$task_type]], predict_type = self$predict_type)
+        fallback_id = mlr_reflections$learner_fallback[[self$task_type]]
+        if (!is.null(fallback_id)) {
+          self$fallback = lrn(mlr_reflections$learner_fallback[[self$task_type]], predict_type = self$predict_type)
```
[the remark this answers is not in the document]
be-marc: > This line assumes that the fallback default fallback learner supports all possible predict types. For classification and regression (once https://github.com/mlr-org/mlr3/pull/1125 is merged) this should be fine. But what about e.g. survival?

We check this now. See https://github.com/mlr-org/mlr3/pull/1151

> Also I think we should not require the is.null(fallback_id) check at runtime. We should just make sure that all the extension packages properly implement this.

It will stay that way for the time being so that we can easily get to CRAN.

sebffischer:
One more comment about the `weights_resampling`. What do you think about not making this weights, but counts instead?
For the holdout resampling, e.g., for a `Task` with row_ids `1:10` we could set the `resampling_weights` to `c(rep(2, 5), rep(1, 5))`.
Then, when sampling the training data we would first modify the `row_ids` to be `c(1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 7, 8, 9, 10)`, then sample the training data and use those IDs that were not sampled for the train set as the test set.

sebffischer:
> One more comment about the `weights_resampling`. What do you think about not making this weights, but counts instead? For the holdout resampling, e.g., for a `Task` with row_ids `1:10` we could set the `resampling_weights` to `c(rep(2, 5), rep(1, 5))`. Then, when sampling the training data we would first modify the `row_ids` to be `c(1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 7, 8, 9, 10)`, then sample the training data and use those IDs that were not sampled for the train set as the test set.

Otoh the question is then whether we really need this because we can already set the `$row_ids` of the task ...

be-marc:
Moved to #1207

PR #1121, opened 2024-08-28
fix: warn instead of error when feature types or levels differ

sebffischer:


be-marc:
Moved to #1127 because force push

PR #1125, opened 2024-08-29
feat: regr.featureless quantile prediction

be-marc:


R/Learner.R
```diff
@@ -572,6 +572,11 @@ Learner = R6Class("Learner",
         fallback_id = mlr_reflections$learner_fallback[[self$task_type]]
         if (!is.null(fallback_id)) {
           self$fallback = lrn(fallback_id, predict_type = self$predict_type)
+
+          if (self$predict_type == "quantiles") {
```
mllg, line 576: Uh, is this really required here in the Learner class?

berndbischl:
see comments in code. otherwise good, fix and merge

R/Learner.R
be-marc: Moved to `set_fallback()` function

PR #1128, opened 2024-08-29
docs: autotest

be-marc:


inst/testthat/helper_autotest.R
```diff
@@ -1,17 +1,37 @@
...
+#' Calls `train()` and `predict()` on the learner and checks the prediction with `score()`.
+#' The prediction is checked with `sanity_check()`.
+#'
+#' `generate_tasks(learner)` generates multiple tasks for a given learner.
+#' Calls `generate_data()` and `generate_generic_tasks()` to generate tasks with different feature types.
+#'
+#' @noRd
+NULL
 
+#' @title Generate Tasks for a Learner
```
berndbischl, line 22: would be nice to add a few lines here a) about the result. i guess it is a list named by properties?
b) can we at least SUMMARIZE what tasks are created? BRIEFLY?

berndbischl:
this is WAY better than before. and pretty good. maybe add the few lines I requested than merge

berndbischl:
see issue #1128, add at least a short entry in dev wiki.

sebffischer:
did you see https://github.com/mlr-org/mlr3/blob/main/R/mlr_test_helpers.R?

inst/testthat/helper_autotest.R
be-marc: Done

be-marc:
> did you see https://github.com/mlr-org/mlr3/blob/main/R/mlr_test_helpers.R?

Both texts reference each other now

PR #1155, opened 2024-08-31
BREAKING CHANGE: encapsulation is not automatically activated when a fallback is set

be-marc:


R/Learner.R
```diff
@@ -566,21 +562,11 @@ Learner = R6Class("Learner",
...
-      # if there is no fallback, we get a default one
-      if (is.null(private$.fallback)) {
-        fallback = default_fallback(self)
-        if (!is.null(fallback)) {
-          self$fallback = fallback
-        }
-      }
     },
 
     #' @field fallback ([Learner])\cr
```
berndbischl, line 567: i would check a little bit more in terms of properties.

if these properties are available in the learner, but NOT the FB I would warn now
2class, multiclass, missings, selected features, 
predict_type the same should be an extra message

berndbischl:
i guess we converge to merging encapsulate and "fallback"

this would mean 

learner$encapsulate(type, fallback)
type: sets enc for both train+pred
fallback: sets the FB, NULL must is possible
if enc is in {callr, evalualte}, FB must be set
if enc = "none", FB must be NULL

learner$fallback stays, but only as a READ ONLY AB.

be-marc:
Closed by #1171

Issue #1198, opened 2024-11-04
benchmark(benchmark_grid(paired=TRUE)) only uses first resampling, ignoring others


tdhock:
related to https://github.com/tdhock/mlr3resampling/issues/19 which instructed me to try `benchmark_grid(paired=TRUE)` (although I would rather not, because I believe paired=FALSE is more user-friendly for my use case).

?benchmark_grid says
```
        * With 'paired' set to 'TRUE', tasks and resamplings are
          treated as pairs. I.e., you must provide as many tasks as
          corresponding instantiated resamplings. The grid will be
```
so I expected that I should provide a list of instantiated resamplings, one for each task, and that each resampling will be used with the corresponding task.

However I observe that only the first item in the resampling list is used.
The problematic line of code is
```r
  if ("resampling" %in% clone) {
    design[, "resampling" := list(list(resampling[[1L]]$clone())), by = list(hashes(resampling))]
  }
```
which has the effect of replicating the first resampling (and ignoring the others).

tdhock:
here is some code to replicate.
```r
x <- 1:100
tsk_1 = mlr3::TaskRegr$new("t1", data.frame(x, y=x+1), target="y")
tsk_2 = mlr3::TaskRegr$new("t2", data.frame(x, y=x+2), target="y")
learner = mlr3::lrn("regr.rpart")
rsmp_cv1 = mlr3resampling::ResamplingVariableSizeTrainCV$new()
rsmp_cv1$instantiate(tsk_1)
rsmp_cv2 = mlr3resampling::ResamplingVariableSizeTrainCV$new()
rsmp_cv2$instantiate(tsk_2)
grid = mlr3::benchmark_grid(list(tsk_1, tsk_2), learner, list(rsmp_cv1, rsmp_cv2), paired = TRUE)
bmr = mlr3::benchmark(grid, store_models = TRUE)
```
I added print statements to benchmark so I can see where the problem is.
```r
> bmr = mlr3::benchmark(grid, store_models = TRUE)
[1] "435b1c2dc90e0299" "9942672582ae8f57"
[1] "435b1c2dc90e0299" "9942672582ae8f57"
[1] "435b1c2dc90e0299" "435b1c2dc90e0299"
Called from: (function (task, learner, resampling, param_values) 
{
    assert_learnable(task, learner)
    if (resampling$task_hash != task$hash) {
        browser()
        stopf("Resampling '%s' was not instantiated with task '%s'", 
            resampling$id, task$id)
[...]

PR #1199, opened 2024-11-05
feat: add task phash

be-marc:
```r
task = tsk("spam")
resampling$instantiate(task)

resample(tsk("pima"), lrn("classif.rpart"), resampling)

#> Error: DataBackend did not return the queried rows correctly: 3067 requested, 501 received
```

Would solve errors like this where the resampling tries to access rows that are not present because the resampling was instantiated on a larger task. In this PR, `benchmark()` and `resample()` check if the resampling was instantiated on the task. This is done with a `Task$phash` that excludes the features.

Would come with an overhead especially in fselect due to the additional hash operations


be-marc:
If we want this, we need to check #1198

be-marc:
The error message has been extended with the sentence "The resampling was probably instantiated on a different task". I will close this PR for now.

Issue #1013, opened 2024-04-07
Callback Hooks for `resample()` / `benchmark()`


sebffischer:
It would give users a lot of flexibility if we would provide some hooks to evaluate additional code during `resample()` and `benchmark()` (and maybe also `batchmark()`. One important use-case is to offer users a replacement for the *holdout task*, which we plan to remove soon. We have already made callbacks available in `mlr3tuning`, `mlr3torch` and `bbotk` etc..
Besides giving users the possibility to define their own callbacks, it would be great if we could also offer predefined callbacks for the most common things a user might want to do. Further, it would be great if we provided convenience functions that enable users to define custom callbacks without knowing too much about R6. `mlr3tuning` and `mlr3torch` also offer something like this. 
Besides evaluating custom code, it should also be possible for these callbacks to store some information in the Result object, e.g. what is returned by a `$get_state()` method.

be-marc:
Closed by #1214

Issue #899, opened 2023-02-10
Possibility to save / access models via callbacks during resample


sebffischer:
In some cases, users might want to fit large models on large datasets using `resample()`.
In this case [here](https://stackoverflow.com/questions/75405217/no-model-stored-with-mlr3), 
the person was unable to fit the models into RAM when specifying `store_models = TRUE`, but he still wanted access to the models, or rather the impurity scores from the models. 

I have to suggestions to make these things more flexible: 
1. Adding callbacks in `resample()` would allow to save these models to disk during `resample()`. 
2. Can we somehow score some measures that depend on the model on the workers? So that measures that require the model can still be evaluated when `store_models = FALSE`?

be-marc:
Same as #1013.

be-marc:
Closed by #1214

PR #1213, opened 2024-11-20
feat: add function to extract information from a model

be-marc:


be-marc:
Closes https://github.com/mlr-org/mlr3/issues/1013 and https://github.com/mlr-org/mlr3/issues/899

be-marc:
```R
task = tsk("iris")

extractor = function(learner) list(selected_features = setdiff(learner$model$frame$var, "<leaf>"))

rr = resample(task, learner, resampling, extractor = extractor)
rr$learners[[1]]$state$selected_features
```

berndbischl:
it is not doced where the result of extract is stored

berndbischl:
also always doc the signature of the function

berndbischl:
also there should be a use case example and test. this should include a benchmark.

a user needs to somehow practilly gain access to the extracted stuff at the end

be-marc:
We will implement #1214  instead.

PR #1214, opened 2024-11-21
feat: add callbacks for resample and benchmark

be-marc:
Additional overhead is less than 3% when using 10ms models

### DIFF
--- a/R/LearnerRegr.R
+++ b/R/LearnerRegr.R
@@ -12,6 +12,7 @@
 #'   - `"se"`: Predicts the standard error for each value of response for each observation in the test set.
 #'   - `"distr"`: Probability distribution as `VectorDistribution` object (requires package `distr6`, available via
 #'     repository \url{https://raphaels1.r-universe.dev}).
+#'  - `"quantiles"`: Predicts quantile estimates for each observation in the test set.
 #'
 #' Predefined learners can be found in the [dictionary][mlr3misc::Dictionary] [mlr_learners].
 #' Essential regression learners can be found in this dictionary after loading \CRANpkg{mlr3learners}.
--- a/man/LearnerRegr.Rd
+++ b/man/LearnerRegr.Rd
@@ -16,6 +16,9 @@ This Learner specializes \link{Learner} for regression problems:
 repository \url{https://raphaels1.r-universe.dev}).
 }
 }
+\itemize{
+\item \code{"quantiles"}: Predicts quantile estimates for each observation in the test set.
+}
 
 Predefined learners can be found in the \link[mlr3misc:Dictionary]{dictionary} \link{mlr_learners}.
 Essential regression learners can be found in this dictionary after loading \CRANpkg{mlr3learners}.


be-marc:
```r
callback = callback_workhorse(
  id = "mlr3.selected_features",
  on_workhorse_before_result = function(callback, context) {
    context$env$learner$state$selected_features = setdiff(context$env$learner$model$frame$var, "<leaf>")
  }
)

task = tsk("iris")

rr = resample(task, learner, resampling, callbacks = callback)

rr$learners[[1]]$state$selected_features
```

R/ContextEvaluation.R
```diff
@@ -0,0 +1,78 @@
...
+    learner = NULL,
+
+    #' @field resampling [Resampling]\cr
+    #' The resampling strategy to be used.
+    #' The resampling is unchanged during the evaluation.
+    resampling = NULL,
+
+    #' @field param_values `list()`\cr
+    #' The parameter values to be used.
+    #' Is usually only set while tuning.
```
sebffischer, line 30: it this not accessible via the learner anyway?

R/ContextEvaluation.R
```diff
@@ -0,0 +1,78 @@
...
+    param_values = NULL,
+
+    #' @field sets (`list()`)\cr
+    #' The train and test set.
+    #' The sets are available on stage `on_evaluation_before_train``.
```
sebffischer, line 35: Maybe instead expose the resampling iter, so we can just get `resampling$test_set(iter)`

R/ContextEvaluation.R
```diff
@@ -0,0 +1,78 @@
...
+    sets = NULL,
+
+    #' @field test_set (`integer()`)\cr
+    #' Validation test set.
+    #' The set is only available when using internal validation.
+    test_set = NULL,
```
sebffischer, line 41: This is accessible via `task$internal_valid_task`, no?

R/as_result_data.R
```diff
@@ -69,6 +80,10 @@ as_result_data = function(task, learners, resampling, iterations, predictions, l
     stopf("Resampling '%s' has not been trained on task '%s', hashes do not match", resampling$id, task$id)
   }
 
+  if (!is.null(data_extra) && length(data_extra) != N) {
+    stopf("Number of data_extra (%i) must match the number of resampling iterations (%i)", length(data_extra), N)
```
sebffischer, line 84:
```suggestion
    stopf("Length of data_extra (%i) must match the number of resampling iterations (%i)", length(data_extra), N)
```

R/benchmark.R
```diff
@@ -201,7 +203,9 @@ benchmark = function(design, store_models = FALSE, store_backends = TRUE, encaps
 
   set(grid, j = "mode", value = NULL)
 
-  result_data = ResultData$new(grid, store_backends = store_backends)
+  data_extra = if (length(callbacks)) map(res, "data_extra")
```
sebffischer, line 206: Do we set `data_extra` to `NULL` if there are callbacks but none of them writes `extra_data`?

R/ContextEvaluation.R
sebffischer: --> I would probably not expose things like this via the context not not restrict ourselves in the future when we want to refactor the workhorse

sebffischer (changes requested):
Looks good! I think we should also provide an example for how to implement a custom callback.

R/CallbackEvaluation.R
```diff
@@ -0,0 +1,123 @@
...
+  )
+)
+
+#' @title Create Workhorse Callback
+#'
+#' @description
+#' Function to create a [CallbackEvaluation].
+#'
+#' ```
+#' Start Evaluation on Worker
```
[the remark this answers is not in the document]
be-marc: Yes, that is more clear. Done.

R/CallbackEvaluation.R
```diff
@@ -0,0 +1,123 @@
...
+    on_evaluation_before_predict,
+    on_evaluation_end ),
+    c(
+      "on_evaluation_begin",
+      "on_evaluation_before_train",
+      "on_evaluation_before_predict",
+      "on_evaluation_end"
+    )), is.null)
+
+  walk(stages, function(stage) assert_function(stage, args = c("callback", "context")))
```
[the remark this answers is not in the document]
be-marc: > I don't think this function works. The different callback stages will have the wrong function environment.

You mean that the complete `callback_evaluation()` function does not work?

> Also we should check that there are no methods that start with on_ that are not in the list of allowed stages. This is likely a typo

Maybe I'm misunderstanding you, but the stages are parameters of `callback_evaluation()` so you can't pass a wrong stage.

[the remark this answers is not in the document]
be-marc: > stages will have the wrong function environment

I just check this. Of course it is the global environment. So it would be good idea to `crate()` these functions?

R/ContextEvaluation.R
be-marc: Yes you are right. Should be enough to make this accessible via `$learner` in `on_evaluation_before_train`.

R/ContextEvaluation.R
be-marc: Yes, when I implemented a callback yesterday, I missed it straight away :D

R/ContextEvaluation.R
be-marc: Yes the only disadvantage of changed `$resampling` and `$task` instead of `$sets` and `$test_set` is that you change these objects also for the next iteration when running this sequentially.

R/benchmark.R
be-marc: No then `data_extra` is created. I thought that `any(map_lgl(res, function(x) is.null(x$data_extra)))` is much more expensive than `length(callbacks)`.  But maybe it is unreasonable to worry about 1/3 milliseconds overhead.

be-marc:
> Looks good! I think we should also provide an example for how to implement a custom callback.

`callback_evaluation()` has an example now

R/ResultData.R
```diff
@@ -68,6 +70,12 @@ ResultData = R6Class("ResultData",
           set(data, j = "resampling", value = NULL)
           set(data, j = "param_values", value = NULL)
 
+          # add extra data to fact table
+          if (!is.null(data_extra)) {
+            assert_list(data_extra, len = nrow(data))
+            set(data, j = "data_extra", value = data_extra)
```
mb706, line 76: list cols can be tricky in data.table, in particular when there are 0 or 1 row tables involved. See

* https://github.com/Rdatatable/data.table/issues/4568
* https://github.com/Rdatatable/data.table/issues/6104
* https://github.com/Rdatatable/data.table/issues/4569

Idk how relevant this is here. E.g. if all tests involve having tables here that have strictly more than 1 row, consider also having a test has  a single row here. But ignore this comment if you know this is not an issue.

R/CallbackEvaluation.R
```diff
@@ -0,0 +1,157 @@
...
+#' @param label (`character(1)`)\cr
+#' Label for the new instance.
+#' @param man (`character(1)`)\cr
+#' String in the format `[pkg]::[topic]` pointing to a manual page for this object.
+#' The referenced help package can be opened via method `$help()`.
+#' @param on_evaluation_begin (`function()`)\cr
+#' Stage called at the beginning of an evaluation.
+#' Called in `workhorse()` (internal).
+#' @param on_evaluation_before_train (`function()`)\cr
+#' Stage called before training the learner.
```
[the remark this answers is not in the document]
be-marc: It no longer seems necessary to have both. The context previously had more objects that were created or modified up to `on_before_train`.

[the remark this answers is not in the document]
be-marc: Maybe it's good to have both. With `on_evaluation_begin`  you could modify the resampling.

R/CallbackEvaluation.R
```diff
@@ -0,0 +1,157 @@
...
+#'
+#' Evaluation callbacks are called at different stages of the resampling process.
+#' Each stage is called once per resampling iteration.
+#' The stages are prefixed with `on_*`.
+#'
+#' ```
+#' Start Resampling Iteration on Worker
+#'  - on_evaluation_begin
+#'  - on_evaluation_before_train
+#'  - on_evaluation_before_predict
```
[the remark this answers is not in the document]
be-marc: Done

R/CallbackEvaluation.R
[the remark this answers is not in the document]
be-marc: Should be clearer now with the new docu

R/ResultData.R
```diff
@@ -18,29 +18,31 @@
 #' print(ResultData$new()$data)
 ResultData = R6Class("ResultData",
   public = list(
+
     #' @field data (`list()`)\cr
-    #'   List of [data.table::data.table()], arranged in a star schema.
```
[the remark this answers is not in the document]
be-marc: I was annoyed that sometimes there are 2 spaces and then again 3 or sometimes none at all. But we can agree on 3 spaces. I also like it more.

be-marc:
Callbacks will behave differently depending on the parallelization. 

```r
  task = tsk("pima")

  callback = callback_evaluation("test",
    on_evaluation_begin = function(callback, context) {
      if (context$iteration == 1) context$task$select(c("age", "insulin"))
    }
  )

  rr = resample(task, learner, resampling, callbacks = callback)

  rr$learners[[1]]$state$feature_names
  # > [1] "age"     "insulin"
  rr$learners[[2]]$state$feature_names
  # > [1] "age"     "insulin"

  with_future("multisession", {
    rr = resample(task, learner, resampling, callbacks = callback)
  })

  rr$learners[[1]]$state$feature_names
  # > [1] "age"     "insulin"
  rr$learners[[2]]$state$feature_names
  # > [1] "age"      "glucose"  "insulin"  "mass"     "pedigree" "pregnant" "pressure" "triceps"
```

I think we should not clone these objects to get a consistent behavior. This will add unnecessary overhead but document this.

R/ResultData.R
be-marc: Thanks for pointing this out. Should be a always a list column now.

R/CallbackEvaluation.R
```diff
@@ -0,0 +1,157 @@
...
+  assert_class(callback, "CallbackEvaluation")
+  invisible(callback)
+}
+
+#' @export
+#' @param callbacks (list of [CallbackEvaluation]).
+#' @rdname assert_evaluation_callback
+assert_evaluation_callbacks = function(callbacks, null_ok = FALSE) {
+  if (null_ok && is.null(callbacks)) return(invisible(NULL))
+  invisible(lapply(callbacks, assert_evaluation_callback))
```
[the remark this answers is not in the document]
be-marc: Fixed

PR #1233, opened 2024-12-20
refactor: predict_type and predict_types

be-marc:


berndbischl:
in the AB: maybe also say in the docs that it is readonly. you probably always want to doc this with the same "keyword".

it could also happen that changing the predict type leads to an error? maybe say that too.

otherwise: fine!

PR #1239, opened 2025-01-01
refactor: use more set_values()

m-muecke:


R/Learner.R
```diff
@@ -75,7 +75,7 @@
 #' To set hyperparameters, assign a named list to the subslot `values`:
 #' ```
 #' lrn = lrn("classif.rpart")
-#' lrn$param_set$values = list(minsplit = 3, cp = 0.01)
+#' lrn$param_set$set_values(minsplit = 3, cp = 0.01)
```
sebffischer, line 78: Here we would also have to change the text around this line. But nice catch!

sebffischer (changes requested):

R/Learner.R
m-muecke: check if you're fine with the new text

sebffischer (approved):

PR #1241, opened 2025-01-06
refactor: simplify dots assertion

m-muecke:
@sebffischer there is also [base::chkDots](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/chkDots), but only throws a warning. Maybe its generic enough to move this assertion to checkmate.


sebffischer:
@m-muecke nice to know, thanks!
Probably makes sense in checkmate too!

sebffischer (approved):

Issue #1270, opened 2025-03-27
`ResampleResult` and `BenchmarkResult` should expose a method to change the threshold after the fact


sebffischer:

sebffischer:
* [ ] Add method `$set_treshold(threshold)` to ResampleResult
* [ ] Add method `$set_threshold(threshold, uhash)` to `BenchmarkResult`
* [ ] Add a way to conveniently retrieve uhashes, by adding `$uhashes` method (with columns, learner_id, task_id, resampling_id, uhash) and documenting this also in `uhash` parameter of `$set_threshold()` and also generally where `uhash` is available as a parameter.

PR #1279, opened 2025-04-09
feat(bmr, rr): Add set_threshold method, uhash_table field and uhash(es) functions

sebffischer:
Resolves Issues #1270

**Summary**

*Threshold Setting*

The objects `ResultData`, `ResampleResult` and `BenchmarkResult` all got methods `$set_threshold()` that allows to change the threshold of probability predictions for binary classification problems.
The only thing that is odd here, is that this method is also available to `BenchmarkResult`s for regression problems. We decided against introducing new specified classes (like `BenchmarkResultClassif`) for now to reduce complexity.

*uhashes*

Selecting for which `ResampleResult`s from a `BenchmarkResult` to change the threshold works via a `uhash`. 
Because other operations (like `$filter()` and `$resample_result()`) operate similarly, we streamlined this as follows:
* `BenchmarkResult` got an AB `$uhash_table` with columns `uhash`, `learner_id`, `task_id` and `resampling_id` which allows to compute specific uhashes from given IDs. To make this more convenient for users, we also added `uhash()` and `uhashes()` methods to do so. These are now also called internally in `$filter()` , `$resample_result()` and `$set_threshold()`, unifying everything.

This is a breaking change because arguments `learner_hashes`, `resampling_hashes` and `task_hashes` were removed from the `Task$filter()` method, but I would expect no user to actually use this interface as it is rather internal.


sebffischer:
TODOs:
* [ ] write out `...` arguments explicitly (`learner_id`, etc.)
* [ ] test  `uhashes()` method with length 0 
* [ ] add parameter `i` to `Task$filter()` method
* [ ] ensure that `uhash` and `i` have the same order in all `Task` methods

PR #1280, opened 2025-04-09
feat(predict-type): Add option for changing default predict type

sebffischer:
**Summary**

* Add a warning when different predict types are specified in learners passed to `benchmark_grid()`.
* Add option to select `"prob"` as the default predict type for classification learners (if available).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -72,6 +72,8 @@ Suggests:
     RhpcBLASctl,
     rpart,
     testthat (>= 3.2.0)
+Remotes:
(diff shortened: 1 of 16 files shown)


sebffischer:
TODOs:
* [ ] use: https://github.com/mlr-org/mlr3misc/pull/133
* [ ] don't add specific flag to disable this warning, make more a general framework (globalCallingHandlers) and document it in mlr3 book: https://github.com/mlr-org/mlr3book/issues/871

Issue #1285, opened 2025-04-15
Calling `task$filter()` with `numeric` argument implicitly converts `task$row_roles$use` to `numeric`


advieser:
With the current dev version of mlr3, most likely due to https://github.com/mlr-org/mlr3/pull/1282, calling `task$filter()` with numeric `row_ids` argument leads to implicit type conversion of `task$row_roles$use` to `numeric` as well. 
```r
library(mlr3)
task = tsk("iris")
class(task$row_roles$use)
# [1] "integer"
task$filter(c(13L, 14L))
class(task$row_roles$use)
# [1] "integer"
task$filter(13)
class(task$row_roles$use)
# [1] "numeric"
```

This leads to errors, e.g. when cbinding:
```r
task$cbind(data.frame(a = 1))
# Error in as_data_backend.data.frame(data, primary_key = pk) : 
#  Assertion on 'data[[primary_key]]' failed: Must be of type 'integer', not 'double'.
```

PR #1286, opened 2025-04-16
fix: implicit type conversion of `task$row_roles$use`

advieser:
closes https://github.com/mlr-org/mlr3/issues/1285


sebffischer:
Thanks @advieser

PR #1299, opened 2025-05-04
Fix: add parameter alpha to pinball measure

lona-k:
The `pinball` function has an additional parameter `alpha` that could not be set for the pinball measure with the `msr()` sugar function.


sebffischer (commented):
I don't think all the cloning is necessary. 
Each time a measure is constructed, a new parameter set is created.

Why was this added?

lona-k:
@sebffischer this wasn't added but copied from the existing code in the `MeasureClassifSimple` class, which already supports adding parameters to simple measures. I just extended the functionality to `MeasureRegrSimple` and `MeasureSimilaritySimple`.

lona-k:
This changes the pinball measure from a `MeasureSimple` into its own measure. This is necessary because the pinball measure needs access to the prediction object in `$score()`, which is not possible with a `MeasureSimple`.

PR #1313, opened 2025-05-21
add ResamplingGroupOrStrata

tdhock:
This PR adds a new class ResamplingGroupOrStrata in between Resampling and the user-facing classes like ResamplingCV.

The idea is that the get_instance() method is defined in ResamplingGroupOrStrata because it is common to user-facing classes like ResamplingCV and ResamplingLOO, but I want to implement a different get_instance() method in mlr3resampling package (while keeping all the other logic of the Resampling class).

@sebffischer
Does that make sense to you?
In combination with https://github.com/tdhock/mlr3resampling/pull/31, it solves https://github.com/tdhock/mlr3resampling/issues/29 for me.
Please review and/or suggest alternatives.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -208,3 +208,4 @@ Collate:
     'task_converters.R'
     'worker.R'
     'zzz.R'
+Remotes: tdhock/mlr3@get_instance
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -208,4 +208,3 @@ Collate:
     'task_converters.R'
     'worker.R'
     'zzz.R'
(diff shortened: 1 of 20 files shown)


sebffischer:
I think this makes sense!
Thanks a lot for the PR.

I would probably: 

1. Make `$.get_instance()` a private method instead
2. Ideally add some documentation (a section "Inheriting") that describes how to inherit from the `Resampling` class, i.e. that it's possible to overwrite both `$.get_instance()` to have full control, or only `.sample()` when one wants to use the predefined mechanism for stratification and grouping.

tdhock:
ok, I made the requested changes, can you please review?

PR #1314, opened 2025-05-22
feat: add mirai parallelization 

be-marc:


inst/testthat/helper_misc.R
```diff
@@ -17,6 +17,14 @@ with_future = function(backend, expr, ...) {
   force(expr)
 }
 
+with_mirai = function(expr) {
+  requireNamespace("mirai")
+  mirai::daemons(1)
+  on.exit(mirai::daemons(0), add = TRUE)
```
mb706, line 23: will this somehow influence mirai sessions that the user has open for another reason?

R/helper_exec.R
```diff
@@ -23,6 +23,9 @@ future_map = function(n, FUN, ..., MoreArgs = list()) {
   if (getOption("mlr3.debug", FALSE)) {
     lg$info("Running experiments sequentially in debug mode with %i iterations", n)
     mapply(FUN, ..., MoreArgs = MoreArgs, SIMPLIFY = FALSE, USE.NAMES = FALSE)
+  } else if (requireNamespace("mirai", quietly = TRUE) && mirai::status()$connections) {
```
mb706, line 26: so this will automatically prefer mirai over future? Maybe there should be an option for this

R/helper_exec.R
```diff
@@ -23,6 +23,9 @@ future_map = function(n, FUN, ..., MoreArgs = list()) {
...
+    lg$debug("Running resample() via mirai with %i iterations", n)
+    mirai::collect_mirai(mirai::mirai_map(data.table(...), FUN, .args = c(MoreArgs, list(is_sequential = FALSE))))
```
mb706, line 28: why does it call `data.table(...)`? And does it have any chance of getting into conflict with any of the special args of `data.table()` (e.g. `key`, `keep.rownames`)

inst/testthat/helper_misc.R
sebffischer: Maybe we can use an `.mlr3` compute profile to avoid this

mb706: sry I see this is only for tests

R/helper_exec.R
sebffischer, line 26: This only works with the default compute profile

sebffischer:
Most importantly: we need to ensure that using `mirai_map` is reproducible and gets proper RNG streams.

R/helper_exec.R
be-marc: Only if the user also started a daemon with `mirai::daemon()`

be-marc: Yes, I didn't want to make it too complicated. What would be the best way to pass the compute profile? We also have to pass it to `mirai_map()` .

inst/testthat/helper_misc.R
```diff
@@ -17,6 +17,14 @@ with_future = function(backend, expr, ...) {
   force(expr)
 }
 
+with_mirai = function(expr) {
+  requireNamespace("mirai")
+  mirai::daemons(1)
```
sebffischer, line 22:
```suggestion
with_mirai = function(expr) {
  with(daemons(1), eval(expr))
}

```

sebffischer:
I 

> > we need to ensure that using mirai_map is reproducible and gets proper RNG streams.
> 
> Yes, for now I pointed out in the documentation that it does not deliver the same results as `future`.
> 
> > One way to achieve this is to use L-Ecuyer-CMRG and generate seeds using
> 
> Yes I did the same in `rush`. Not sure why Charlie is doing it differently.

Actually, Charlie is also doing it. But every mirai worker instantites the L-Ecuyer seed once at startup.

But this causes e.g. the following behavior: 

``` r
library(mirai)

set.seed(1)
daemons(2)
#> [1] 2
x1 = mirai_map(1, function(i) rnorm(1))[][[1]]
set.seed(1)
x2 = mirai_map(1, function(i) rnorm(1))[][[1]]
x1 == x2
#> [1] FALSE
```

<sup>Created on 2025-05-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I think it would be more user friendly for us to also set the seeds again before starting the parallel work.

sebffischer:
Another idea for optimization:

If we had more control over the daemons, e.g. via a special compute profile `.mlr3`, we could also globally load some packages on the workers (e.g. `mlr3` itself). 
Not sure whether results in a noticeable runtime improvement but we could explore it.

R/helper_exec.R
sebffischer: I suggest we expose a `.compute` argument to `benchmark()` and `resample()`.

be-marc: I'm not so happy with that because we are introducing parameters for a special backend. Moreover, this argument would then have to be added to all tuning and feature selection functions.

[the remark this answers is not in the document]
be-marc: Done

be-marc:
> Actually, Charlie is also doing it. But every mirai worker instantites the L-Ecuyer seed once at startup and `set.seed()` then has no effect on the parallel workers.

Does this mean that the substreams are not attached to the tasks but to the daemons? It somehow sounds like that in https://mirai.r-lib.org/reference/daemons.html#arg-seed. Otherwise, the order in which tasks are sent would have no influence.

sebffischer:
> > Actually, Charlie is also doing it. But every mirai worker instantites the L-Ecuyer seed once at startup and `set.seed()` then has no effect on the parallel workers.
> 
> Does this mean that the substreams are not attached to the tasks but to the daemons? It somehow sounds like that in https://mirai.r-lib.org/reference/daemons.html#arg-seed. Otherwise, the order in which tasks are sent would have no influence.

Yes, I think so

R/helper_exec.R
shikokuchuo: I wouldn't worry about this. Already in a `with(daemons(...), { })` scope `mirai()` calls will default to those settings with `.compute = NULL`. More to come on this front.

be-marc:
The results are now reproducible with mirai 2.4.1 if a seed is passed to `miria::daemons()`. However, the results are not reproducible across parallel backends.

man-roxygen/section_parallelization.R
```diff
@@ -1,8 +1,11 @@
...
-#' This function can be parallelized with the \CRANpkg{future} package.
-#' One job is one resampling iteration, and all jobs are send to an apply function
-#' from \CRANpkg{future.apply} in a single batch.
+#' This function can be parallelized with the \CRANpkg{future} or \CRANpkg{mirai} package.
+#' One job is one resampling iteration.
+#' All jobs are send to an apply function from \CRANpkg{future.apply} or `mirai::mirai_map()` in a single batch.
 #' To select a parallel backend, use [future::plan()].
+#' To use `mirai`, call `mirai::daemons()` before calling this function.
+#' The `future` package guarantees reproducible results independent of the parallel backend.
+#' The results of `mirai` will not be the same but can be made reproducible by setting a `seed` and `dispatcher = FALSE` when calling `mirai::daemons()`.
```
sebffischer, line 9: this is outdated

R/Learner.R
```diff
@@ -539,6 +539,11 @@ Learner = R6Class("Learner",
     #' * `"callr"`: Uses the package \CRANpkg{callr} to call the learner, measure time and do the logging.
     #'   This encapsulation spawns a separate R session in which the learner is called.
     #'   While this comes with a considerable overhead, it also guards your session from being teared down by segfaults.
+    #' * `"mirai"`: Uses the package \CRANpkg{mirai} to call the learner, measure time and do the logging.
+    #'   This encapsulation calls the function in a `mirai` on a `daemon`.
+    #'   The `daemon` can be pre-started via `daemons(1)`, otherwise a new R session will be created for each encapsulated call.
+    #'   If a `deamon` is already running, it will be used to executed all calls.
+    #'   Using `mirai"` is similarly safe as `callr` but much faster if several learners are encapsulated one after the other on the same daemon.
```
mllg, line 546: What about reproducibility?

R/helper_exec.R
```diff
@@ -23,6 +23,9 @@ future_map = function(n, FUN, ..., MoreArgs = list()) {
   if (getOption("mlr3.debug", FALSE)) {
     lg$info("Running experiments sequentially in debug mode with %i iterations", n)
     mapply(FUN, ..., MoreArgs = MoreArgs, SIMPLIFY = FALSE, USE.NAMES = FALSE)
+  } else if (requireNamespace("mirai", quietly = TRUE) && mirai::daemons_set()) {
```
mllg, line 26: `isNamespaceLoaded("mirai")`

R/helper_exec.R
mllg: As discussed today: to avoid larger objects to be serialized for every row, use multiple calls to `mirai_map()` + `.args`

mllg (approved):

PR #1332, opened 2025-07-05
Fix autotest `configure_learner`

bblodfon:
see https://github.com/mlr-org/mlr3extralearners/issues/381

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mlr3
 Title: Machine Learning in R - Next Generation
-Version: 1.0.0.9000
+Version: 1.0.1.9000
 Authors@R:
   c(
     person("Michel", "Lang", , "michellang@gmail.com", role = "aut",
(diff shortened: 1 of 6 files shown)


tests/testthat/test_autotest.R
```diff
@@ -41,4 +41,23 @@ test_that("autotest recognizes learner that does not use .get_weights()", {
...
+test_that("autotest configure_learner works", {
+  task = tsk("spam")
+  learner = lrn("classif.debug", error_train = 1)
+  # keep only feat_all_* tasks to save time
+  exclude = "^feat_single|^missings|^weights|^utf8|^sanity"
+  result = run_autotest(learner, 20, exclude)
+  expect_false(result$ok)
+
+  # example changes the learner's parameter during the autotest
+  cfg_lrn = function(learner, task) {
```
sebffischer, line 55: if this is the `configure_learner` function, shouldn't the test check that `error_train = 0` in one case?

PR #1339, opened 2025-07-14
fix: sprintf in autotest

AnnaNzrv:
### Problem description
This PR fixes an issue in the autotest (`helper_autotest`) where error messages were constructed using `sprintf(info, ...)` within the `err()` function. This caused tests to fail when error messages contained unescaped "%" characters (e.g., from matrix multiplication or other operations that return messages with "%" symbols).

### Proposed fix 
Remove "%" characters from the info string (= error message) before passing it to `sprintf()`.


be-marc:
Thanks!

Issue #450, opened 2020-03-04
Properly test sparse backend


mllg:
As a proof of concept. Also implement `fix_factor_levels.Matrix()`.

mllg:
Fixed via #1363.

PR #1363, opened 2025-08-11
Remove deprecated data formats and Matrix backend

mllg:
This includes:

* argument `data_format`
* Learner field `data_formats` 
* `DataBackendMatrix`

DataBackendMatrix has been around since mlr3's start, but still no learner makes use of it. With only a single supported format (`data.frame`/`data.table`), the already deprecated support for different data formats can be dropped.


mllg:
Fixes #450

mllg:
Quite surprisingly, `DataBackendMatrix` was not mentioned once in the book. 😟

PR #1389, opened 2025-09-07
chore(partition): make partition a S3 generic

m-muecke:
* use explicit integers
* use implicit return
* else is sufficient since checking for length in assertion


be-marc:
Thanks!

PR #1418, opened 2025-11-30
Implement deadline feature

advieser:


advieser:
Tests failed previously due to https://github.com/r-lib/processx/issues/402.

Two open points to discuss:
* Types: Right now we would also allow numerics which we'd then internally convert to POSIXct. This is mainly to allow `$deadline[["train"]] = Inf` instead of `$deadline[["train"]] = as.POSIXct(Inf)`. I'd be in favor of being more strict concerning types.
* Names: @mb706 argued against allowing merging of values, as is done for `$timeout` (see lines below), because the AB should be treated like the named vector it is. A consequence of this would be that the user must always pass a named vector length 2 to `$configure()`, for example. At the same time, this would disallow `$deadline <- c(train = Sys.time(), train = Sys.time() + 1)`. 
We should probably agree on one solution or the other for both timeout and deadline before merging.
https://github.com/mlr-org/mlr3/blob/ae1543e3645fb200dd32dd6646ca4da38d41821b/R/Learner.R#L1057-L1058

mb706:
Agree that we should be strict and only allow `as.POSIXct(Inf)`

R/Learner.R
```diff
@@ -1059,6 +1060,23 @@ Learner = R6Class(
...
+    #' Default is `as.POSIXct(c(train = Inf, predict = Inf))`.
+    deadline = function(rhs) {
+      if (missing(rhs)) {
+        return(private$.deadline)
+      }
+      if (is.numeric(rhs)) {
+        rhs = as.POSIXct(rhs)
+      }
+      assert_posixct(rhs, any.missing = FALSE, len = 2L, .var.name = "deadline")
+      assert_names(names(rhs), subset.of = c("train", "predict"), .var.name = "names of deadline")
```
be-marc, line 1076:
```suggestion
      assert_names(names(rhs), type = "unique", subset.of = c("train", "predict"), .var.name = "names of deadline")
```

be-marc: Please also fix for `$timeout`

R/Learner.R
```diff
@@ -1059,6 +1060,23 @@ Learner = R6Class(
...
+
+      private$.deadline = rhs
```
be-marc, line 1078:
```suggestion
      private$.deadline[names(rhs)] = rhs
```

be-marc: Allows partial updates

R/Learner.R
```diff
@@ -1059,6 +1060,23 @@ Learner = R6Class(
       private$.timeout
     },
 
+    #' @field deadline (named `POSIXct(2)`)\cr
+    #' Deadline for the learner's train and predict steps. This uses the same mechanism as `timeout`, but with absolute
```
be-marc: One sentence per line

be-marc (changes requested):

PR #1433, opened 2026-01-21
docs: consistent type annotations for NULL

be-marc:


cmdanurag (changes requested):
I noticed that `MeasureInternalValidScore` has a nullable parameter (`select = NULL`) where the type annotation currently states `character(1)` but does not include `| NULL`.

It does not seem to be covered in this PR. Would you like me to extend this PR to include it?

be-marc:
> Would you like me to extend this PR to include it?

Yes

be-marc:
@cmdanurag Open a new PR if you still want to work on this

PR #1511, opened 2026-09-18
fix: include timeout, deadline, and encapsulation in Learner$hash

be-marc:
## Summary

`Learner$hash` and `$phash` excluded `timeout`, `deadline`, the encapsulation method, and the `when` handler. `benchmark()` groups the learners of the design by hash and replaces each group by a clone of its first member, and `ResultData` stores one learner object per hash. Two learners differing only in those settings were therefore silently merged: the second one ran with the first one's settings, and `bmr$learners` reported only the first.

## Repro on main

```r
l1 = lrn("classif.rpart"); l2 = lrn("classif.rpart"); l2$timeout = c(train = 5)
l1$hash == l2$hash   # TRUE
design = data.table(task = list(task, task), learner = list(l1, l2), resampling = list(rs, rs))
benchmark(design)    # both rows run with timeout Inf, result holds one learner
```

## Fix

Add `.timeout`, `.deadline`, `.encapsulation`, and `.when` to both `hash` and `phash` and document them. The fallback was already part of the hash, so the encapsulation settings belong there as well. `calculate_hash()` hashes the `when` function by its formals and body, so equal handlers hash equal.

This changes all learner hashes, noted in NEWS. `GraphLearner` in mlr3pipelines overrides `hash`/`phash` and needs the same fields in a follow-up there.

Tests: hash and phash change for each of the four settings and stay equal for identical configuration; a benchmark with two learners differing only in `timeout` runs each with its own timeout (checked via a callback) and reports both learners.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


be-marc:
Follow-up for `GraphLearner`, which overrides `hash`/`phash`: mlr-org/mlr3pipelines#1033

